PR #190, opened 2015-01-05
Orchestration API update

magixyu:
1. StackUpdate supports Environment file and Template file with "get_file"
2. Add findStack API.

### DIFF
--- a/core/src/main/java/org/openstack4j/api/heat/StackService.java
+++ b/core/src/main/java/org/openstack4j/api/heat/StackService.java
@@ -55,6 +55,14 @@ public interface StackService {
 			boolean disableRollback, Long timeOutMins);
 
 	/**
+     * returns details of a {@link Stack}.
+     * @param stackName
+     *            Name of {@link Stack}
+     * @return {@link Stack}
+     */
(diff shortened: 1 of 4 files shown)


magixyu:
```java
Stack myStack = os.heat().stacks().findStack("mystack");
os.heat().stacks().update("isc02", myStack.getId(), Builders.stackUpdate()
                                          .templateFromFile("template.yaml)
                                          .environmentFromFile("env.yaml")
                                          .timeoutMins(5).build());
```

gondor:
Awesome Thank you!  FYI,  I ended up renaming findStack to be getStackByName.  This follows inline with the current pattern with OpenStack4j when finding some object/entity by it's name.  For example UserServer.getByName("user") is just one example.

magixyu:
That's ok. I use "findStack" just copy from the openstack document :)

gondor:
No problem.  Early on when I first started this project I decided on consistent naming conventions.  Multiple results is always a "list" call and a single result is a "get" variant.  Just makes it nice jumping between the various services as a new user and having common patterns.

Issue #183, opened 2014-12-18
Heat: Nested templates (Files & Env) support for a Stack


maishsk:
Is it possible to call nested templates? I could not find how to do it.
Thanks

gondor:
Support was added

PR #196, opened 2015-01-08
Heat: Nested templates (Files & Env) support for a Stack #183

magixyu:
Add the Nested template support to resolve issue 183.


gondor:
Awesome!

PR #212, opened 2015-01-22
support lbaas V1.0. test in juno installed by devstack.

liujunpengwork:
pool: list/get/create/update/associate/dissociate/delete
vip:list/get/create/update/delete
member:list/get/create/update/delete
healthmonitor:list/get/create/update/delete


gondor:
@liujunpengwork  Wow this looks great! Thank you!  I'm going to make one quick change after I merge.  I going to add a loadbalancers() interface from the networking() service and then add all the new service interfaces under that.  

For example: os.networking().loadbalancers().pool() etc.

PR #230, opened 2015-01-29
 Allow to get all images without setting the "limit" parameter.

magixyu:
Allow to get all images without setting the "limit" parameter.
It looks that the openstack has default PAGE_SIZE for the api request.
The /images/details cannot return all, in my environement, only 25.


gondor:
Looks great and good observation and solution.  Thank you!

PR #231, opened 2015-01-29
Add the function to update the query parameters

magixyu:
Add the function to update the http query parameters for which only has one value not a list.
Like when using query parameter "marker" to query the image list, the "marker" should be updated to use the last image id from the response for next request, not appending the value.


gondor:
Looks good!

Issue #234, opened 2015-01-30
Volumes cannot be used on HP Cloud


akanto:
Hi,

The Volume management the Openstack4J fails on HP Cloud:
 
```
Exception in thread "main" org.openstack4j.api.exceptions.RegionEndpointNotFoundException: Endpoint for 'cinder' service could not be found for region: 'region-a.geo-1'
```
I believe the problem is that type of BlockStorage is 'volume' on HP Cloud and Openstack4J expects 'cinder'.
```
Service:AccessService{name=Block Storage, type=volume, endpoints=[KeystoneEndpoint{region=region-a.geo-1, publicURL=https://region-a.geo-1.block.hpcloudsvc.com/v1/11362286233782, versionId=1.0, versionInfo=https://region-a.geo-1.block.hpcloudsvc.com/v1, versionList=https://region-a.geo-1.block.hpcloudsvc.com}, KeystoneEndpoint{region=region-b.geo-1, publicURL=https://region-b.geo-1.block.hpcloudsvc.com/v1/11362286233782, versionId=1.0, versionInfo=https://region-b.geo-1.block.hpcloudsvc.com/v1, versionList=https://region-b.geo-1.block.hpcloudsvc.com}]}
```

Kind regards,
Attila

akanto:
Hi,

Pull request has been created: https://github.com/gondor/openstack4j/pull/235

Kind regards,
Attila

gondor:
Oh, I already fixed this.  Was just deploying a new SNAPSHOT.  Will checkin my fix, looks like the same as yours

akanto:
Many thanks. It is working :).

PR #235, opened 2015-01-30
#234 checking a list of ServiceType values when resolving the URL

akanto:
The Volume management the Openstack4J fails on HP Cloud, because the Block Storage type is volume on HP Cloud and not cinder.


gondor:
Fixed issue with same fix

Issue #214, opened 2015-01-22
Add OSGi support to JARs


logoff:
It would be very useful adding OSGi support in project JARs. Now it is not possible adding project JARs (and dependencies) into and OSGi platform (like [Apache Felix](http://felix.apache.org/) or [Eclipse Equinox](http://eclipse.org/equinox/)).

Adding support to OSGI is very easy using [Apache Felix Maven Bundle Plugin](http://felix.apache.org/site/apache-felix-maven-bundle-plugin-bnd.html).

Finally, creating a [Apache Karaf Feature](http://karaf.apache.org/manual/latest/users-guide/provisioning.html) would ease the job of adding the project as dependency in [Apache Karaf](http://karaf.apache.org/).

Some of the project dependencies are OSGi compatible but there is a mess in their dependency chain (import and export relations are not correct at all!).

gondor:
I would love to have support for OSGI in OpenStack4j.  In my engineering path I've haven't had an environment that leveraged it, so I'm not that versed in it.  Is this something you'd want to contribute?  If not I may have more questions if I look into this.

isartcanyameres:
Hi, I'm currently working on this feature.

The steps I'm following:
- use [Apache Felix Maven Bundle Plugin](https://felix.apache.org/site/apache-felix-maven-bundle-plugin-bnd.html) to create OSGi bundles during project build (it only changes the MANIFEST.MF in generated jar)
- use [Aries SPI Fly](http://aries.apache.org/modules/spi-fly.html#specconf) required headers to make SPI work in OSGi.

I'll rise any question I may have using the [openstack4j group](https://groups.google.com/forum/#!forum/openstack4j)

gondor:
Great, thank you!  I just saw your question in the group.  I'll provide a few answers and as well I can call out who recently did the yaml work so they can respond on those questions.

isartcanyameres:
Although previous patches prepare openstack4j to be depoyed in OSGi environments, currently **openstack4j is not yet functional in OSGi**.

Currently, openstack4j-core cannot be resolved in OSGI as it makes use of a package that is not visible in OSGi environments. 

TODO: 
Find an alternative implementation for org.openstack4j.openstack.heat.utils.Environment.java and org.openstack4j.openstack.heat.utils.Template.java without importing packages that are not visible in OSGi. Please see the [conversation about using not visible package](https://groups.google.com/forum/#!topic/openstack4j/k3W5xn4L7zo)

[...]

PR #242, opened 2015-02-03
Feature/osgi compatible

isartcanyameres:
This patch adds OSGi compatibility to openstack4j. It is related with issue #214

It uses maven-bundle-plugin to generate the bundle manifest of each project and then includes it in the jar.

This patch also provides a karaf feature to easily deploy openstack4j in this platform. Current karaf feature include only openstack4j-core with openstack4j-http and openstack4j-log4j. If other deployments are desired, new features may be added to this file at will.

CONSIDERATIONS
openstack4j uses SPI. In order to support SPI in OSGi this patch makes use of [Aries SPI Fly](http://aries.apache.org/modules/spi-fly.html). Bundles using SPI should register as
providers and/or consumers, as explained [here](http://aries.apache.org/modules/spi-fly.html#specconf)
It turns out that due to this registration, consumer bundles are not active until all required services are satisfied by at least one provider. This means that default provider mechanism used in openstack4j-core with **FallbackLoggerFactorySupplier does not work in OSGi**.

Currently, **openstack4j is not yet functional in OSGi**.
openstack4j-core cannot be resolved in OSGI as it makes use of a package that is not visible in OSGi environments. Please see the [conversation about using not visible package](https://groups.google.com/forum/#!topic/openstack4j/k3W5xn4L7zo)

TODO: 
Find an alternative implementation for org.openstack4j.openstack.heat.utils.Environment.java and org.openstack4j.openstack.heat.utils.Template.java without importing packages that are not visible.


gondor:
Awesome work Isart!  Thank you!

PR #252, opened 2015-02-25
add new function for setting adminPassword when launch the server

frank-zera:


gondor:
Looks good! Thank you

Issue #255, opened 2015-02-27
Need a connector with java.net.HttpURLConnection


krishnabrucelee:
The current set of connectors depend on 3rd party libraries and the dependency chain grows when we need a small implementation. Having a plain implementation with java.net.HttpURLConnection will help for OSGi based apps.

gondor:
@krishnabrucelee thank you for the contribution. I have reviewed your PR and added some comments.  Once you address (re-commit fixes) / answer them I should be able to merge this.  Thanks again!

PR #256, opened 2015-02-27
Added new connector module using HttpURLConnection

krishnabrucelee:
Fix for issue #255 
Added the connector with HTTPUrlConnection


connectors/http-connector/pom.xml
```diff
@@ -0,0 +1,12 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.pacesys.openstack4j.connectors</groupId>
+        <artifactId>openstack4j-connectors</artifactId>
+        <version>2.0.2-SNAPSHOT</version>
+    </parent>
+    <name>OpenStack4j HttpURL Connector</name>
+    <artifactId>openstack4j-http-connector</artifactId>
+    <packaging>jar</packaging>
```
gondor: Travis CI is failing and local tests too since the maven bundle plugin wasn't added.  Look at another connector and copy in the <build> .. </build> tag replacing the package info

connectors/http-connector/src/main/java/org/openstack4j/connectors/http/HttpCommand.java
```diff
@@ -0,0 +1,190 @@
...
+                out.write(requestBody.toString().getBytes());
+                out.flush();
+            }
+
+            in = new BufferedReader(new InputStreamReader(
+                    connection.getInputStream()));
+            String inputLine;
+
+            while ((inputLine = in.readLine()) != null) {
+                contentBuilder.append(inputLine);
```
gondor: I'm a little worried about this since we are taking the input stream and converting it to a String.  I have a feeling Swift (S3) will be broken streaming files back down, etc.   Could we pass the input stream to the Response object so if OS4j calls the connectors response getInputStream for binary downloads and in cases where like readyEntity is called you could apply the conversion to a String above within the response object.

connectors/http-connector/src/main/java/org/openstack4j/connectors/http/HttpResponseImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(HttpResponseImpl.class);
+
+    private Map<String, List<String>> headers;
+    private int responseCode;
+    private String responseMessage;
+    private String body;
+
+    private HttpResponseImpl(Map<String, List<String>> headers,
+            int responseCode, String responseMessage, String body) {
```
gondor: Instead of String body I'm suggesting the raw input stream

connectors/http-connector/src/main/java/org/openstack4j/connectors/http/HttpResponseImpl.java
```diff
@@ -0,0 +1,139 @@
...
+                retHeaders.put(key, value);
+            }
+        }
+
+        return retHeaders;
+    }
+
+    @Override
+    public <T> T readEntity(Class<T> typeToReadAs) {
+        try {
```
gondor: This is the method that could take the inputStream and apply the logic you had in HttpCommand to read as a String

connectors/http-connector/src/main/java/org/openstack4j/connectors/http/HttpResponseImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public String getStatusMessage() {
+        return responseMessage;
+    }
+
+    /**
+     * @return the input stream
+     */
+    public InputStream getInputStream() {
+        return null;
```
gondor: Since I'm suggesting creating the Response with the input stream this could simple return it now

gondor:
@krishnabrucelee Thank you and overall looks really good!  Made some suggestions since I think a couple of our API services may fail with your String approach.  See comments and let me know if you need further clarification.

connectors/http-connector/pom.xml
krishnabrucelee: I faced connection time out issue in all the connectors including HTTPURLConnector.

krishnabrucelee:
Travis CI has passed all the connectors and local tests. We are awaiting your response.

gondor:
This looks great! Thank you for adding!

PR #271, opened 2015-03-14
Project model class for tenants in Keystone v3

krishnabrucelee:
Project Model class use to group/isolate resources and/or identity objects. Project is the new name for tenants in Keystone v3.


gondor:
Rejecting pull request due to it being incomplete.  Project is not utilized anywhere in the stack.  Changes to the POM corrupt integrity of our current 2.0.2 branch.

PR #275, opened 2015-03-14
Added java file NeutronExternalGatewayInfo

MonaCIET:


gondor:
Without this being tied into to any services or referenced by anything else it currently has no benefit in this project at this point.

MonaCIET:
I have added ExternalGatewayInfo to return router object inorder to update the service class.

Issue #258, opened 2015-03-02
reset vm state support


octupszhang:
I suggest that openstack4j should give  support for reseting vm state. It's useful to repair vm  when vm change to error state because of the failures of migration,resize...

gondor:
@octupszhang Ok support added to the server service.  Checked into master branch and a new 2.0.2-SNAPSHOT deployed.

Usage:
```java
ActionResponse ar = os.compute().servers().resetState(serverId, Status.ACTIVE);
```

octupszhang:
Thanks for your help.

PR #283, opened 2015-03-17
change state type in ResetAction from enum to String

octupszhang:
#258 when i use resetStateAction, openstack returns error like:
default ,I change ResetStateAction.state form enmu to string, and it works well now.I'm not sure wheather it's hard code, just for reference.


gondor:
I'm going to keep this open and investigate first.   There should be no reason why I couldn't keep the enum vs a string so this is essentially a bug.  If there is no way I'll merge this.  Thanks and this is good to see as a working example.

gondor:
@octupszhang I just made the fix which eliminates the need for the string.  Please review against Issue #285 and let me know if this solves it for you.

PR #287, opened 2015-03-18
Added create and delete ceilometer alarms functionality

e3ky:
Added create and delete ceilometer alarms functionality.
Few files are a little modified. One new file is introduced.


gondor:
Thanks for the contribution! Approved

PR #309, opened 2015-03-31
Add ability to consume trust on authentication

peter-nordquist:
Added some code to the V3 Auth builder so it can consume a trust when authenticating against Keystone


peter-nordquist:
Well I don't understand github well enough so I'm going to re-create this with branches for each pull

PR #311, opened 2015-03-31
Pull in user info from V3 Authentication

peter-nordquist:
I needed the user information, specifically user_id for my workflow using the V3 APIs.  This is a little dangerous since I moved the Domain specific classes out of the Project specific classes since Domains apply to Projects and Users so anyone that has imported them will need to re-import.


gondor:
Change makes sense and I agree with doing it now to support future V3 work.

PR #315, opened 2015-03-31
Use code from http-connector to fix issues with Swift (Object Store) put

peter-nordquist:
Using payloads and the Swift API in OpenStack4j was erroring out at the Jackson level since it could not serialize InputStreams to JSON.  I've pulled in code from the http-connector to handle InputStreams similarly in the okhttp-connector.


gondor:
Great catch! Thanks.  The connectors were all recently released so are still in beta stages.  Appreciate the extra testing and helping the stabilizing efforts.

Issue #343, opened 2015-04-23
Add filter support when listing Volume and Volume Snapshot


bsemp:
Add filter support when listing Volume and Volume Snapshot
Openstack Cinder supports filters in `list` function.
Openstack4J does not currently provide this functionality.

PR #344, opened 2015-04-23
Add support for filter when listing volumes and volume snapshots

bsemp:
Issue #343


gondor:
Awesome! Thanks Bruno

PR #351, opened 2015-04-27
Exposing Firewall Rules for a Firewall Policy.

vishvesh:
1) Now exposing NeutronFirewallRules (removed previous JsonIgnore).
2) There seems to be some problem with `Lists.newArrayList()` while
fetching Rules for a policy.. replaced with `new
ArrayList<NeutronFirewallRule>()`.

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/networking/domain/ext/NeutronFirewallPolicy.java
+++ b/core/src/main/java/org/openstack4j/openstack/networking/domain/ext/NeutronFirewallPolicy.java
@@ -1,5 +1,6 @@
 package org.openstack4j.openstack.networking.domain.ext;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import org.openstack4j.api.Apis;
@@ -13,7 +14,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonRootName;
 import com.google.common.base.Objects;
-import com.google.common.collect.Lists;
 
 /**
  * A Neutron Firewall (FwaaS) : Firewall Policy Entity.
@@ -42,7 +42,6 @@ public class NeutronFirewallPolicy implements FirewallPolicy {
 	@JsonProperty("firewall_rules")
 	private List<String> firewallRules;
 	
-	@JsonIgnore
 	private List<NeutronFirewallRule> neutronFirewallRules;
 		
 	/**
@@ -100,7 +99,7 @@ public class NeutronFirewallPolicy implements FirewallPolicy {
 	@Override
 	public List<? extends FirewallRule> getNeutronFirewallRules() {
         if (neutronFirewallRules == null && (firewallRules != null && firewallRules.size() > 0)) {
-        	neutronFirewallRules = Lists.newArrayList();
+        	neutronFirewallRules = new ArrayList<NeutronFirewallRule>();
             for (String ruleId : firewallRules) {
                 NeutronFirewallRule rule = (NeutronFirewallRule) Apis.getNetworkingServices().
                 		firewalls().firewallrule().get(ruleId);


gondor:
@vishvesh Could you look at the tests you've added.  There failing due to some JSON

vishvesh:
Oh is it? Let me check..

gondor:
From what I am reading above you have a couple options.  I've done this lots in the past by adding a static class extending the original.  

Take a look at https://github.com/gondor/openstack4j/blob/632cfc3692ee811ce0b0e465414b9c1edf30119d/core/src/main/java/org/openstack4j/openstack/compute/domain/NovaQuotaSet.java 

Scroll down to the bottom and you can see how I changed the root name in that class yet the contents are the same as the parent.  

So you have a couple options.  

1.  You could experiment first by creating a static class like I did with the QuotaSet and use that one for the removal.  Do not put the @JsonRootName in this case.  I think this will work so may want to try this first.  Looking at the ObjectMapperSingleton I'm always looking at the top most class vs iterating down the hierarchy looking for the @JSonRootName annotation

2. If #1 fails you could rename the NeutronFirewallPolicy to something like AbstractNeutronFirewallPolicy and then create two new classes extending this (one for each use case).  One would contain the @JsonRootName and the other without.  

Jeremy

vishvesh:
Was going for #2 currently :)

Will try #1 as you mentioned, test it and get back to you. Thanks!

vishvesh:
@gondor : #1 did not seem to work. I went with approach #2, and have tested it thoroughly now.
Everything seems to be working as expected (with some changes). I have created a pull request #353 for the same.

Question: When I do a `get()` operation on `FirewallPolicy`, I am expecting a call to fetch me firewall rules associated with the policy --> `getNeutronFirewallRules()` - which was the original intention of this commit. But instead, it is not calling. If you get some time, could you take a look at `AbstractNeutronFirewallPolicy --> getNeutronFirewallRules()` call and suggest if I'm missing any annotation?

vishvesh:
@gondor : When I do a `get()` operation on `FirewallPolicy`, I am expecting a call to eager fetch firewall rules associated with the policy --> `getNeutronFirewallRules()` - which was the original intention of this commit. It is eager fetching the rules associated as expected. Buy it is also trying to push the field `neutronFirewallRules` while creating a Firewall Policy - which is not expected, hence failing the creation.  If you get some time, could you take a look at `AbstractNeutronFirewallPolicy --> getNeutronFirewallRules()` call and suggest if I'm missing any annotation?

The request that it is making is the following: 
```
POST http://xx.xx.xx.xx:9696/v2.0/fw/firewall_policies
{
  "firewall_policy" : {
    "name" : "new11",
    "description" : "new11",
    "shared" : false,
    "audited" : false,
    "neutronFirewallRules" : [ {
      "id" : "0957cf64-215a-420f-8a56-abe183278a74",
      "name" : "new1",
      "description" : "new1",
      "enabled" : true,
      "shared" : false,
      "action" : "allow",
      "protocol" : "tcp",
      "tenant_id" : "7069831a4bdc485ea0696988b7af81ad",
      "ip_version" : 4
    } ],
    "firewall_rules" : [ "0957cf64-215a-420f-8a56-abe183278a74" ]
  }
}
```
Basically, it is first fetching the rule associated, & then putting it in `neutronFirewallRules` field. All we expect is `firewall_rules` while creation, and `neutronFirewallRules` while fetching the policy.

P.S. I tried adding annotation on setter & annotation on the field, but it doesn't work: 
```
@JsonIgnore
public void setNeutronFirewallRules(List<NeutronFirewallRule> neutronFirewallRules) {
	this.neutronFirewallRules = neutronFirewallRules;
}
```
I have also compared `NeutronNetwork` class, and it pretty much looks the same for `neutronSubnets`. Could you suggest what's going on? Thanks!

gondor:
Jackson typically never cares about the setter.  But it will look at the getter.  Did you try @JsonIgnore on the `getter` and `field`?

vishvesh:
Yes I have tried both.. Not sure how it's working for `NeutronNetwork` and not for `neutronFirewallRules `..

gondor:
So I ran some quick tests and by adding:

```java
@JsonIgnore
	@Override
	public List<? extends FirewallRule> getNeutronFirewallRules()
```
It stopped including the neutronFirewallRules: [ ] in the create call. That was the only spot I had to add the JsonIgnore.  Maybe my test wasn't the same as yours.

vishvesh:
@gondor : I have tried that.. and the behavior you mentioned is expected.. Problem with this is, it doesn't `eager` fetch us the `firewallRules` while invoking `get()` operation..

gondor:
Oh I see what you mean.  Why does it need to?  I don't believe the NeutronNetwork is either.

vishvesh:
While doing `get()` or `list()` operation on NeutronNetwork, I see call to `neutronSubnets`, and it fetches the subnets associated to that network.. I thought I'd do the same for convenience/saving 1 more call to fetch the rule entity by id..

gondor:
Ok, the big difference is in NeutronNetwork we don't allow you to addSubnets during create (not in the spec).  In your scenario you can add rules on create via passing the Id's of each rule.  Unfortunately on the POST your "neutronRules" is null and your firewallRules (id version) has values which triggers the method to lazy fetch based on the current logic.  

To get around all of this you could keep the firewallRules (ie. List<String>) in the Abstract as `protected`.

I would then move the getNeutronFirewallRules into your NeutronFirewallPolicyRule class.   You should then create a NeutronFirewallPolicyRuleCreate extending from Abstract and have your concrete builder actually user the NeutronFirewallPolicyRuleCreate class.  This way it will not have the eager getNeutronFirewallRules and no more @JsonIgnore's will be needed.    Hope that makes sense.

vishvesh:
Ohh I see.. Thanks for the detailed explanation :+1: I'll probably end up making 1 more call to fetch the associated rules or keep the firewallPolicy as a `managed` entity and call `getNeutronFirewallRules()`.. Should I just commit with `@JsonIgnore` added on `getNeutronFirewallRules()` call?
```
@JsonIgnore
@Override
public List<? extends FirewallRule> getNeutronFirewallRules()
```

gondor:
I don't think it matters at this point.  Might need to do a few tests.  Btw, when querying for firewall rules do you know if policy aka policyId is a filterable param?  If so it might be more efficient to just call the firewall rules service using the `list` operation and a filter of the current policy.  Saves all the calls to the server if thats possible since you can do a single call to retrieve a list using the current policy as a filter source.  

Just thinking out loud.

vishvesh:
Yes, it is filterable afaik. I was thinking the same. I don't like making too many ajax calls to get a logical entity presented to the user anyway. I first have to fetch all the policies & then I'll make use of the filterable param option or just call `getNeutronFirewallRules()` :) Thanks for the time & discussion.. I'll make sure I test with the JsonIgnore & commit it tomorrow.. :+1:

PR #353, opened 2015-04-28
Changes in NeutronFirewallPolicy.

vishvesh:
1) Separated NeutronFirewallPolicy by creating parent
AbstractNeutronFirewallPolicy class which is extended by
NeutronFirewallPolicy & NeutronFirewalPolicyRule classes.
	     --> Prior has `@JsonRootName("firewall_policy")` attribute whereas the
later doesn't have (which is used by `rule_insert/rule_remove` calls -
which doesn't require JsonRootName - `firewall_policy`).
2) Fixed and added `firewallpolicyrule` test cases to add/remove rules
from a policy.
3) Removed null checks for insertRuleInPolicy call for `type` and
`insertAfterOrBeforeRuleId` fields.
4) Added `firewall_list` field in AbstractNeutronFirewallPolicy as this
extra field is returned by OpenStack after `add/remove rule` calls.
5) User can now invoke `insertFirewallRuleInPolicy` with null parameters.
```
FirewallPolicy ruleAdded =  osClient.networking().firewalls().firewallpolicy()
						.insertFirewallRuleInPolicy(policyId, ruleId, null, null);
```


gondor:
Awesome! thanks

Issue #360, opened 2015-05-06
Connection leak when using HttpClient connector


bsemp:
Since issue #332 , when using HttpClient connector with some method  like `blockStorage().snapshots().delete`, connections seems to be not properly cleaned after request, resulting in HttpClient PoolingHttpConnectionManager waiting indefinitely for a slot to become available again.

The issue seems to target functions returning `ActionResponse` object like 'delete' as the 'get' and 'list' methods are not affected.

Currently, the only workaround is to increase both the parameters _MaxConnectionsPerRoute_ and _MaxConnections_ but this cannot be a viable solution for long running application.

Find below a quick test demonstrating the issue:

```bash
for COUNT in {1..10}; do
    cinder snapshot-create --display-name test-$COUNT xxxxxxxxxxxxxx
done
```

```java
public class TestHttpClientConnectionLeak {

    private static final String ENDPOINT_URL = "xxxxxxxxxxxxxx";
    private static final String USERNAME = "xxxxxxxxxxxxxx";
    private static final String PASSWORD = "xxxxxxxxxxxxxx";
    private static final String TENANT_NAME = "xxxxxxxxxxxxxx";
    
    private OSClient client;
    
    /**
     * @param args the command line arguments
     */
    public static void main(String[] args) throws Exception {
        TestHttpClientConnectionLeak test = new TestHttpClientConnectionLeak();
        
        //HttpLoggingFilter.toggleLogging(true);
        
        test.init();
        test.runGetTest();
        test.runDeleteTest();
    }
    
    private void init() {
        client = OSFactory.builder()
                .endpoint(ENDPOINT_URL)
                .credentials(USERNAME, PASSWORD)
                .tenantName(TENANT_NAME)
                .perspective(Facing.PUBLIC)
                .withConfig(Config.newConfig())
                .authenticate();
    }
       
  private void runGetTest() throws Exception {
        for (int i = 1; i <= 10; i++) {
            VolumeSnapshot snap = getVolumeSnapshotByName("test-"+i);
            final String id = (snap != null)? snap.getId():"xxx";
            System.out.println("Getting test-" + i + ": " + id);
            client.blockStorage().snapshots().get(id);
        }
    }

    private void runDeleteTest() throws Exception {        
        for (int i = 1; i <= 10; i++) {
            VolumeSnapshot snap = getVolumeSnapshotByName("test-"+i);
            final String id = (snap != null)? snap.getId():"xxx";
[...]

PR #362, opened 2015-05-07
Fix connection leak when mapping HttpResponse to ActionResponse.class…

bsemp:
… or Void.class when using HttpClient

As the response content is not read, the connection is not automatically closed: explicitly add a close call.

Fix file indentation

Issue #360


gondor:
Thank you for providing the fix.  I've been working on another request and haven't had time to run your tests for the root cause.

PR #372, opened 2015-05-12
Add scope when authenticating with an auth token

erikringsmuth:
This seems to cover issue https://github.com/gondor/openstack4j/issues/366. There is only 1 place that uses this constructor so I think we can modify the existing one. Let me know what you think.


gondor:
This looks good! Thank you for the change

PR #379, opened 2015-05-19
Add an annotation to easily skip tests for one or more connectors

bsemp:
Sample:
@Test
@SkipTest(connector = "Http URL Connector", issue = 999, description = "a description")
public void f() {...}


gondor:
Great Idea!!

PR #386, opened 2015-05-25
endpoint resolution for Keystone V3 catalog services now includes region

petergardfjall:
The patch submitted in this pull requests fixes an issue I discovered with openstack4j when used in a multi-regional OpenStack deployment.

More specifically, the code in its current state does not include the region specified via OSClient.useRegion() when matching endpoints from a  Keystone V3 service catalog (I'm not sure if the same issue exists when using Keystone version 2). 

The proposed patch makes sure that the DefaultEndpointURLResolver selects a service endpoint from the right region (if a region has been specified in the URLResolverParams).


gondor:
Looks good! Thank you

PR #437, opened 2015-08-14
add support to list available zone list

csm1sh:
there is requirement to support list available zone list


csm1sh:
add support to list available cinder list

PR #438, opened 2015-08-18
Add support for RAX API key authentication extension

frsyuki:
Rackspace supports API-key based authentication mechanism documented at
http://docs.rackspace.com/files/api/v1/cf-devguide/content/Retrieving_Auth_Token.html
and http://docs.rackspace.com/openstack-extensions/auth/RAX-KSKEY-admin-devguide/content/RackExt-0001.html.

We can use API key instead of password + tenantName with this mechanism.

This is useful especially when we want to improve security using multi-factor
authentication because password-based authentication fails if
multi-factor authentication is enabled.

We can use this extension with this code:

```java
client = OSFactory.builder()
    .endpoint("https://identity.api.rackspacecloud.com/v2.0")
    .credentials("myuser", "mayapikey")
    .raxApiKey(true)
    .authenticate()
client.objectStorage().objects().list("mycontainer")
```

I implemented it as a part of V2 authentication because it's an extension of authentication V2.

### DIFF
--- a/core/src/main/java/org/openstack4j/api/client/IOSClientBuilder.java
+++ b/core/src/main/java/org/openstack4j/api/client/IOSClientBuilder.java
@@ -96,6 +96,14 @@ public interface IOSClientBuilder<R, T extends IOSClientBuilder<R, T>> {
         V2 tenantId(String tenantId);
         
         /**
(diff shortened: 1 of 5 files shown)


frsyuki:
Thank you for merging! I created another pull-request: #439.
We can migrate away from jclouds with these 2 changes.

PR #451, opened 2015-09-10
Add capability to modify ConfigDrive value

### LINKED ISSUE #450: Make configDrive() accessible from NovaServerCreate.ServerCreateConcreteBuilder
Hi!

I need to set configDrive to true, but the Builder presently don't allow to do so !
https://github.com/gondor/openstack4j/blob/5f0d6648949cc3b869ac99265a5971e7526caa6d/core/src/main/java/org/openstack4j/openstack/compute/domain/NovaServerCreate.java

Can you add this capability, or if there's an other solution point me to it ?

Thanks in advance!

imclem:
Fix #450


imclem:
Any update here ?

PR #457, opened 2015-09-17
Correction nested template

draoullig:
Hi,

I change the name of the template push to the files map. In fact if the name is relative, the mapping between the master template and the sub template doesn't match because the name isn't found in the map because it became an absolute name.

Bests regards.


gondor:
Awesome, thank you. Will merge and deploy a new snapshot

PR #466, opened 2015-10-02
Do not throw Exceptions in static initializer

olivergondza:
When openstack4j fails to load `HttpExecutorService` implementation it throws an exception while initializing class (static field `HttpExecutor.INSTANCE` -> constructor -> `throw`). The real cause is wrapped in unhelpful exception:

```
java.lang.NoClassDefFoundError: Could not initialize class org.openstack4j.core.transport.internal.HttpExecutor
	at org.openstack4j.openstack.internal.OSAuthenticator.authenticateV2(OSAuthenticator.java:97)
	at org.openstack4j.openstack.internal.OSAuthenticator.invoke(OSAuthenticator.java:48)
	at org.openstack4j.openstack.client.OSClientBuilder$ClientV2.authenticate(OSClientBuilder.java:107)
	at org.openstack4j.openstack.client.OSClientBuilder$ClientV2.authenticate(OSClientBuilder.java:71)
```

This is what user get after the fix:

```
org.openstack4j.api.exceptions.ConnectorNotFoundException: No OpenStack4j connector found in classpath
	at org.openstack4j.core.transport.internal.HttpExecutor.service(HttpExecutor.java:31)
	at org.openstack4j.core.transport.internal.HttpExecutor.execute(HttpExecutor.java:49)
	at org.openstack4j.openstack.internal.OSAuthenticator.authenticateV2(OSAuthenticator.java:97)
	at org.openstack4j.openstack.internal.OSAuthenticator.invoke(OSAuthenticator.java:48)
	at org.openstack4j.openstack.client.OSClientBuilder$ClientV2.authenticate(OSClientBuilder.java:107)
	at org.openstack4j.openstack.client.OSClientBuilder$ClientV2.authenticate(OSClientBuilder.java:71)
```

PR #467, opened 2015-10-02
Kill context classloader

olivergondza:
`ServiceLoader.load(Class)` relies on context class loader by default. As openstack4j does not provide any, this effectively delegates the responsibility for resetting context classloaders with each call that looks up extensions (which is a process that should be encapsulated) to the client.

I am using openstack4j in environment where context classloaders can not load necessary classes so without this fix, I have to wrap the calls into:

```java
        final Thread thread = Thread.currentThread();
        final ClassLoader oldLoader = thread.getContextClassLoader();
        thread.setContextClassLoader(getClass().getClassLoader());
        try {
            // Now use the convenient openstack4j DSL
        } finally {
            thread.setContextClassLoader(oldLoader);
        }
```

I suggest to rely on whatever classloader loaded openstack4j classes. If this turns out to be necessary to configure, clean API for providing classloades should be created.

Without this fix I am getting exceptions like:
```
java.lang.NoClassDefFoundError: Could not initialize class org.openstack4j.api.Apis
	at org.openstack4j.openstack.internal.OSClientSession.networking(OSClientSession.java:260)
```

CC #466

### DIFF
--- a/core/src/main/java/org/openstack4j/api/Apis.java
+++ b/core/src/main/java/org/openstack4j/api/Apis.java
@@ -82,7 +82,7 @@ public class Apis {
 	}
 	
 	private static APIProvider initializeProvider() {
-		APIProvider p = ServiceLoader.load(APIProvider.class).iterator().next();
+		APIProvider p = ServiceLoader.load(APIProvider.class, Apis.class.getClassLoader()).iterator().next();
 		p.initialize();
 		return p;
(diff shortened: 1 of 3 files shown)


gondor:
I like this.  Thanks for catching it and pointing it out.  Merging and will re-deploy a new snapshot shortly.

Issue #497, opened 2015-11-05
Domain scoped authentication


auhlig:
Currently our identity v3 initial implementation #480 enables OpenStack4j only to authenticate with project scope. 
This issue addresses the implementation of missing domain scoped authentication.
Includes tests.

PR #503, opened 2015-11-09
domain scoped authentication with tests

auhlig:
This PR related to issue Domain scoped authentication #497 and enables OpenStack4j to authenticate with a domain scope in the identity v3 case.
This is the last commit for this kind of issues which is compatible to the v2 implementation.

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/identity/domain/v3/AccessWrapper.java
+++ b/core/src/main/java/org/openstack4j/openstack/identity/domain/v3/AccessWrapper.java
@@ -14,6 +14,7 @@ import org.openstack4j.model.identity.Role;
 import org.openstack4j.model.identity.Tenant;
 import org.openstack4j.model.identity.TokenV2;
 import org.openstack4j.model.identity.v3.Catalog;
+import org.openstack4j.model.identity.v3.Domain;
 import org.openstack4j.model.identity.v3.Project;
 import org.openstack4j.model.identity.v3.User;
 import org.openstack4j.openstack.identity.domain.KeystoneTenant;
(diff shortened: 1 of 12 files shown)


auhlig:
Hey @mazimo, 
Would you mind taking a look?
BR

core/src/main/java/org/openstack4j/openstack/identity/domain/v3/AccessWrapper.java
```diff
@@ -63,17 +64,40 @@ public static AccessWrapper wrap(KeystoneToken token) {
...
+            Tenant tenant = KeystoneTenant.builder()
+                    .id(project.getId())
+                    .name(project.getName())
+                    .description(project.getDescription())
+                    .enabled(project.isEnabled())
+                    .build();
+
+            return new V2Token(id, token.getExpires(), token.getVersion(), tenant);
+        }
+        else if(token.getDomain() != null) {
```
gondor: Probably could break this out of a nested else sequence. ie.  if (token.get...)   since the previous condition is returning a token

gondor:
Looks good @auhlig from an implementation standpoint.  LGTM

auhlig:
If there aren't any objections or comments by @mazimo ,maybe we could merge and include this in the next release?

core/src/main/java/org/openstack4j/openstack/identity/domain/v3/AccessWrapper.java
auhlig: Right. But if you don't mind I would leave it that way for now and do better on the new dedicated v3 branch :)

gondor:
:+1: LGTM

Issue #498, opened 2015-11-05
Add initial implementation of V3 UserService


auhlig:
This issue is for the initial implementation of the user service for identity v3 covering all CRUD operations.
Comes with tests.

auhlig:
Already done & tested :) 
Push postponed after tomorrows brainstorming/discussion on issue #495.

Issue #495, opened 2015-11-05
Create new OSClient for Identity V3-based authentication


dhague:
The existing OSClient interface is V2-specific, having a declaration:
`Access getAccess();`
This maps to the result of V2-based authentication.

PR #516, opened 2015-11-26
OpenStack4j Identity V3

auhlig:
This PR comes with new initial v3-..

OSClient (clean v3 interfaces) #495
IdentityService
UserService #498
Authentication (domain-, project-, no- scope, regions..) #497
Connectors supporting PATCH-method
Mock-tests to that #498
Updated readme.md (how to)

Future PR's will definitely address less issues at once.

This is a proposal for solving those issues and hopefully a good basis for a discussion.


auhlig:
Hey @gondor, @mazimo, @dhague would you mind taking a look?

dhague:
+1 LGTM

gondor:
:+1: LGTM

mhshams:
@gondor / @auhlig / @dhague, 

I am using this branch (v3_collaboration) to test authentication against keystone v3. although everything seems ok for v3 but  using this branch I no longer able to authenticate against v2. Is this something temporary or you are going to drop v2 support in this version permanently?

dhague:
@mhshams The plan is to drop v2 permanently for OpenStack4J 3.x - users who wish to continue with v2 authentication should use OpenStack4J 2.x

mhshams:
@dhague  thanks for the quick reply. 
I am not sure about the effort, but supporting both v2 and v3, in the same ops4j version, would be very useful. I think keystone v2 is going to be around at least for a couple of years and convincing providers to support v3 wouldn't be easy, at least in my case. 
So for the time being I need to bundle two versions of ops4j (v2.x and v3.x) in my software to be able to support keystone v2 and v3 which is possible, thanks to osgi, but it is not an ideal solution.

mazimo:
@gondor / @auhlig / @dhague / @mhshams 

I'm not sure we should drop support for v2 in OpenStack4J. By all means, add v3; but please continue to support v2 in the latest releases.

Imagine software that depends upon OpenStack4J to manage multiple OpenStacks (e.g. from many different data centers); some of the OpenStacks only expose v2 identity endpoints in the service catalog, some only expose v3 identity endpoints. If OpenStack4J provides support for both v2 and v3, then only one dependency is required to OpenStack4J--it can be the latest stable release--and the fluent java API maintained (e.g. identity.v2.authenticate(...) / identity.v3.authenticate(...). 

If v2 is dropped, this becomes somewhat painful to achieve... What do you think?

Additional: I imagine that there will be many data centers that continue with v2 for the next couple of years at least, even though many are currently moving to v3.

PR #524, opened 2015-12-01
moved grant/revoke/check-role from user- to role-service

auhlig:
Moved the grant/revoke/check-user-role for domain and project context from the userService to the roleService. That would follow the v2-pattern of where this kind of calls should live.
Any comments by @gondor, @dhague ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,6 @@ OSClient os = OSFactory.builder()
 ```
 (2) authenticate with domain-scope
 ```java
-// Identity V2 Authentication Example
 OSClient os = OSFactory.builder()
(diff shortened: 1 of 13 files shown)


dhague:
Apart from my comment above (which is not strictly related to this refactoring), LGTM.

gondor:
Outstanding work!  LGTM!

auhlig:
We found a bug in OpenStack that seems to affect every connector except for Jersey2 as described in #530.

But I added failure checks for grant/revoke userrole in domain and project context.

dhague:
Good stuff

Issue #567, opened 2016-02-01
"Mapping image to local is not supported" error is returned by OpenStack when boot index is specified


akanto:
Hi, 

The boot_index is serialised as string by openstack4j, therefore starting from Kilo version we are getting back a "Mapping image to local is not supported" from OpenStack when block device mapping is specified.

```
{
  "server" : {
    "name" : "test",
     ...
    "block_device_mapping_v2" : [ {
     ...
      "boot_index" : "0",
      "delete_on_termination" : true
    } ]
  }
}
```

The boot_index shall be serialised as integer.

Thanks.

PR #568, opened 2016-02-01
convert boot_index to integer in JSON

akanto:
This PR solves #567. The root cause is that OpenStack check boot_index == 0:

```
if not (image_uuid_specified and boot_index == 0):
                  raise exception.InvalidBDMFormat(
                         details=_("Mapping image to local is not supported."))
```

https://github.com/openstack/nova/commit/cadbcc440a2fcfb8532f38111999a06557fbafc2#diff-16c3690d30123cdcd3a1dca53021513aR201


gondor:
Looks good!   Could this potentially affect users of Juno?

akanto:
We are using Juno and Kilo, and tested against both and works fine.

gondor:
Awesome, I'll merge and publish a snapshot this afternoon.  Thanks for the contrib.

PR #570, opened 2016-02-10
Use Travis CI container-based infrastructure

gonzolino:
Remove workaround for Travis CI bug https://github.com/travis-ci/travis-ci/issues/5227 and disable sudo to ensure builds run on the travis container-based infratsructure, where the bug described in above link should not occur.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,6 @@ jdk:
   - oraclejdk7
   - openjdk7
 
-# Workaround for https://github.com/travis-ci/travis-ci/issues/5227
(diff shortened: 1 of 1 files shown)


gonzolino:
Seems like that didn't work...
I will investigate the problem tomorrow.

gonzolino:
Setting a short hostname using the new hostname addon from Travis CI did not work.
I just discovered that you can force the build to run on a docker container by setting `sudo: false` in .travis.yml.
Travis container-based infrastructure does not suffer from the long hostnames, and therefore makes the workaround for short hostnames obsolete.

PR #584, opened 2016-02-27
Embeded jackson-dataformat-yaml as a non-OSGi jar dependency

hyunsun:
Fixed unresolved constraint issue when bundling openstack4j in OSGi environment by embedding jackson-dataformat-yaml as a non-OSGi jar dependency.


gondor:
Awesome, thank you!

Issue #544, opened 2016-01-18
Add initial implementation of V3 RoleService


auhlig:
This issue is for the initial implementation of the role service for identity v3 covering all CRUD adn role assignment operations.
Comes with tests.

PR #590, opened 2016-03-02
identity v3 role service with tests

auhlig:
Brings the V3 Role Service covering CRUD and role assignment operations as described in #544. Comes with tests.

What do you think @gondor, @dhague ?


gondor:
Wow! Looks great @auhlig .

dhague:
+1 Looks good

gondor:
Merging!

Issue #545, opened 2016-01-18
Add initial implementation of V3 Service and Endpoint Service API


auhlig:
This issue is for the initial implementation of the identity v3 service, which handles services and endpoints covering all CRUD operations.
Comes with tests.

PR #592, opened 2016-03-04
identity v3 service and service endpoints

auhlig:
This PR adds the service and service endpoints CRUD operations as described in #545. 

What do you think @gondor, @dhague ?


gondor:
Looks good!

dhague:
Looks good to me

PR #599, opened 2016-03-07
Added noGateway method to the subnet builder

fmarinelli:
Using noGateway method will build a slightly different NeutronSubnet, the only one difference is the property gateway_ip annotated with @JsonInclude to force serialization also if the value is null.
This change will let you add a subnet with gateway disabled.

### DIFF
--- a/core/src/main/java/org/openstack4j/model/network/builder/SubnetBuilder.java
+++ b/core/src/main/java/org/openstack4j/model/network/builder/SubnetBuilder.java
@@ -68,6 +68,11 @@ public interface SubnetBuilder extends Builder<SubnetBuilder, Subnet> {
     SubnetBuilder gateway(String gateway);
 
     /**
+     * @see Subnet#isNoGateway()
+     */
+    SubnetBuilder noGateway();
+
+    /**
      * @see Subnet#getDnsNames()
      */
 	SubnetBuilder addDNSNameServer(String host);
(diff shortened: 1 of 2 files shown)


gondor:
LGTM! Thanks

ayberkcal:
I have faced the same problem. "gateway_ip" property is not included in request json.
Subnet Builder Code:
```
Subnet subnet=getSubnetById(subnetId); //getting subnet from OS
SubnetBuilder builder=subnet.toBuilder(); //create builder from existed subnet
builder=builder.noGateway(); //disable subnet gateway
openStackService.client().networking().subnet().update(builder.build()); //send request to OpenStack API
```

Debug Logs:

> >> PUT /v2.0/subnets/9180b203-d4cb-4c92-9d3c-5b9958b10b82 HTTP/1.1 
> >> "{[\r][\n]" 
> >> "  "subnet" : {[\r][\n]" 
> >> "    "name" : "subnet name123",[\r][\n]" 
> >> "    "dns_nameservers" : [ ],[\r][\n]" 
> >> "    "allocation_pools" : [ {[\r][\n]" 
> >> "      "start" : "192.164.1.2",[\r][\n]" 
> >> "      "end" : "192.164.1.254"[\r][\n]" 
> >> "    } ],[\r][\n]" 
> >> "    "host_routes" : [ ][\r][\n]" 
> >> "  }[\r][\n]" 
> >> "}" 

As you can see "gateway_ip" property is not converted to json. I have debug code. Everything looks fine but samething went wrong when entity converting to json.

Issue #602, opened 2016-03-10
ERROR: Root name 'quotas' does not match expected ('quota')


ivokotev:
Current implementation of class NeutronNetQuota maps at JsonRootName("quota"). However current Network Extension API returns array of "quotas" - http://developer.openstack.org/api-ref-networking-v2-ext.html#listQuotas

This leads to the following exception:
`[org.openstack4j.connectors.httpclient.HttpResponseImpl] ERROR: Root name 'quotas' does not match expected ('quota') for type [simple type, class org.openstack4j.openstack.networking.domain.NeutronNetQuota]
 at [Source: org.apache.http.conn.EofSensorInputStream@c0b41d6; line: 1, column: 2]`

PR #603, opened 2016-03-10
Fix for issue #602

ivokotev:
ERROR: Root name 'quotas' does not match expected ('quota')

Change mapping of JSON from 'quota' to 'quotas' according to Current Network Extensions API
http://developer.openstack.org/api-ref-networking-v2-ext.html#listQuotas


gondor:
Thanks, however  this was not the correct solution which is why tests are failing.  You need to wrapper the root.  I just checked in a fix.  Closing this PR.

Issue #579, opened 2016-02-24
Problem in connection pooling using httpclient as connector


sjpatil:
We are using 2.10 release with httpclient and facing issue with connection pool.

We are doing multiple server snapshot downloads in a sequence. First download works well but later downloads get stuck. As by default the max connections per route is 2; when increased withMaxConnectionsPerRoute , we could do more successful calls. But ultimately it gets stuck again after the connections are exhausted. It seems connections are open and not getting closed.

   Config config=Config.newConfig().withMaxConnectionsPerRoute(4)

Logs where it gets stuck are -

 2016/02/23 08:47:41:573 UTC [DEBUG] PoolingHttpClientConnectionManager - Connection [id: 0][route: {}->http://endpoint:5000] can be kept alive indefinitely
2016/02/23 08:47:41:573 UTC [DEBUG] PoolingHttpClientConnectionManager - Connection released: [id: 0][route: {}->http://endpoint:5000][total kept alive: 2; route allocated: 1 of 2; total allocated: 4 of 20]
 2016/02/23 08:47:41:574 UTC [DEBUG] RequestAddCookies - CookieSpec selected: best-match
 2016/02/23 08:47:41:574 UTC [DEBUG] RequestAuthCache - Auth cache not set in the context
 2016/02/23 08:47:41:574 UTC [DEBUG] PoolingHttpClientConnectionManager - Connection request: [route: {}->http://endpoint:8774][total kept alive: 2; route allocated: 2 of 2; total allocated: 4 of 20]

Our download code makes sure that the input stream that we receive is closed. However, all operations from ServerService get stuck after the last successful download, however other calls from ImageService can be made.

vinodborole:
@gondor Can you please help us here?

sachinwalia2k8:
@vinodborole is this still an issue?

dbantchovski:
@vinodborole 

Yes, this is still an issue. I am facing a similar problem /http connection leak when using apache http client connector/. My scenario is complex, but I managed to extract the problematic part from it:

`Server server = os.compute().servers().get(serverId);
os.compute().servers().createSnapshot(serverId, "one");
os.compute().servers().createSnapshot(serverId, "two");
os.compute().servers().createSnapshot(serverId, "three");`

The third invocation hangs, waiting to obtain a connection - /not that by default apache http client uses pool of 20, connections, but only 2 per host/:

java.lang.Thread.State: WAITING (parking)
[...]

PR #604, opened 2016-03-11
A fix regarding connection leak problem #579

dbantchovski:
In case of http call with response, the returned entity should be consumed in order to release the resources associated with the call, and especially the http connection.

I am not sure where is the right place to consume and release the entity. I made a fix for my use-case - create snapshot, but most probably the this fix is applicable for more situations.


gschukin:
@dbantchovski you mean in httpclient? I've faced something similar developing connector for httpclient5, but since it's not GA didn't take a deep look.

gondor:
Great fix! Thank you

dbantchovski:
@gondor 
Thank you too. I am wondering ... in which version this fix will be included and when it will be released?

gondor:
I'll cut another release this weekend so it will be in 2.20

Issue #540, opened 2016-01-08
Identity V3 token service


auhlig:
This issue is for missing methods as specified in http://developer.openstack.org/api-ref-identity-v3.html#authenticate-v3 like checkToken(), revokeToken(), ..
Comes with tests.

PR #609, opened 2016-03-17
identity v3 token service

auhlig:
This addresses #540 and brings the missing token- check(), validate() and revoke() methods.

Note: This PR introduces the last service for the identity v3 part. The next PR should be the final before one before the release which will focus on the v2/v3 compatibility.

What do you think @gondor and @dhague ?


dhague:
Looks good

gondor:
Looks great!

PR #611, opened 2016-03-18
Move ActionResponse into common package

### LINKED ISSUE #610: ActionResponse class in wrong package
The ActionResonse class is used in most (if not all) API service classes.
Still, the class is located in the org.openstack4j.model.compute package.

The class should be moved to the common package.

gonzolino:
Currently the ActionResponse class is part of the
org.openstack4j.model.compute package. As it is used in most
(if not all) API service classes, it should be moved into the
org.openstack4j.model.common package.

This PR resolved #610


gondor:
@gonzolino We are almost about to release v3 of OpenStack4j (currently different branch) and work by @auhlig @dhague .  Can you send a pull request to ```v3_collaboration``` branch for this as well.  We need to keep things in sync

gonzolino:
@gondor ok i will do that over the weekend.
The patch doesn't merge automatically into the v3_collaboration branch, so i will have to rebase it.

PR #621, opened 2016-03-24
Move ActionResponse into common package for v3

### LINKED ISSUE #610: ActionResponse class in wrong package

gonzolino:
Currently the ActionResponse class is part of the
org.openstack4j.model.compute package. As it is used in most
(if not all) API service classes, it should be moved into the
org.openstack4j.model.common package.

This PR resolves #610 

This is a rebased version of #611 targeted for the v3_collaboration branch


auhlig:
:+1:

gondor:
@auhlig fyi, letting you control merges into the V3 branch :)

auhlig:
Thanks :)

PR #626, opened 2016-03-30
Fixed #625 getByName failing when result list is empty 

### LINKED ISSUE #625: ArrayIndexOutofBoundException for getByName() in V3 Identity Service
In V3.0

Minor issue in the following methods in the case when the User, Project or Service does not exist.

org.openstack4j.openstack.identity.internal.GroupServiceImpl.getByName(String, String)
org.openstack4j.openstack.identity.internal.ProjectServiceImpl.getByName(String, String)
org.openstack4j.openstack.identity.internal.UserServiceImpl.getByName(String, String)

Will throw an ArrayIndexOutofBoundException.

Following the API of other services, I suggest that the Identity services return the entity (User,Project or Service) or null if not found

anneraby:
Added Test Cases for failing getByName in 

org.openstack4j.openstack.identity.internal.GroupServiceImpl.getByName(String, String)
org.openstack4j.openstack.identity.internal.ProjectServiceImpl.getByName(String, String)
org.openstack4j.openstack.identity.internal.UserServiceImpl.getByName(String, String)

And fixed the problem by adding a first() method in ListResult<T> and calling it from the respective implementations.

Let me know what you think.

### DIFF
--- a/core-integration-test/src/test/groovy/org/openstack4j/api/identity/KeystoneGroupServiceSpec.groovy
+++ b/core-integration-test/src/test/groovy/org/openstack4j/api/identity/KeystoneGroupServiceSpec.groovy
@@ -197,6 +197,12 @@ class KeystoneGroupServiceSpec extends AbstractSpec {
         //
         //		then: "we should no longer find the group used in this scenario"
         //		groupList.contains(group) == false
+		
+		when: "non-existent group is listed by name and domain id"
+		Group nonExistent_group_byName_byDomainId = os.identity().groups().getByName("nonExistentGroup", DOMAIN_ID)
+
+		then: "the return value should be null"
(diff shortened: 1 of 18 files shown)


auhlig:
Hi Anne,

Nice work. :+1: 

As we did the majority of the identity v3 tests in groovy using spock framework and betamax to record and replay client-server transactions in the core-integration-test module would it be much work to add the tests there too? So we could have the full picture there. For example: https://github.com/gondor/openstack4j/blob/v3_collaboration/core-integration-test/src/test/groovy/org/openstack4j/api/identity/KeystoneGroupServiceSpec.groovy

anneraby:
Worked in the missing tests cases in Groovy and Spock.

auhlig:
Cool. Thanks for the additional work. Looks good from my side.

PR #627, opened 2016-04-01
the successfull response state of updateMetadata is 202 instead of 204

YanjunSun:
see: http://developer.openstack.org/api-ref-objectstorage-v1.html


YanjunSun:
the successfull response state of updateMetadata is 202 instead of 204

gondor:
thanks for catching this

PR #630, opened 2016-04-06
Refactoring and cleanup before release

auhlig:
As described in detail in issue #629 this PR brings the identity v2 and v3 world together in one version of OS4j.

After that one is merged I think we're good to release.

What do you think @gondor and @dhague ?

PR #562, opened 2016-01-27
Switch to SLF4J as only logging implementation

gschukin:
This change is not backward compatible with earlier version.

Starting now it is required to specify proper slf4j implementation eg logback or log4j2 otherwise slf4j will fallback to NOP logger

slf4j and log4j plugins also were removed as they are obsolete now

PR #632, opened 2016-04-12
slf4j as sole logging impl. for v3_collaboration

auhlig:
I agree with the proposal from #630 to make slf4j the only logging implementation.
This is a modified version of #562 for v3_collaboration.

Would you mind taking a look @gondor and @dhague ?


gondor:
So sorry I missed this @auhlig . LGTM, thanks

dhague:
Looks good - much tighter code as a result.

auhlig:
Cool. Sorry for the delay. Merging now.

Issue #496, opened 2015-11-05
Use Java 7 as the minimum version for openstack4j v3


dhague:
openstack4j 2.x uses Java 1.6 as the minimum version according to pom.xml.

1.6 was EOL in Feb 2013 (see http://www.oracle.com/technetwork/java/eol-135779.html), and openstack4j 3.x will be based on the Identity V3 API, which was released in OpenStack Grizzly in April 2013.

Therefore it makes sense to have a minimum of at least Java 7 for the v3 release, and this would allow us to use the additional features introduced in Java 7 (see http://www.oracle.com/technetwork/java/javase/jdk7-relnotes-418459.html).

gondor:
Good call.  I agree with this change.  As part of this change could you also remove the jersey2-jdk16 connector/module. It was simply there for JDK16 users at the time but going forward won't be needed.

Issue #563, opened 2016-01-27
Update jackson to 2.7 for v3


gschukin:
Update jackson from 2.4 to 2.7

sachinwalia2k8:
@gondor @gschukin - if jackson is upgraded to 2.7 then JDK 6 may be used as a dependency. However 2.8 onwards all support for JDK 6 will be dropped. Is there a timeline until when you want to support JDK 6?

gondor:
@sachinwalia2k8 , @auhlig @dhague have been working on the new seamless V3 integration in a v3_collaboration branch.  This will end up being merged in and bump OpenStack4j from the 2.x to 3.0.0 release.   At that time the ongoing work will be in OpenStack4j version 3 (new master) which has also dropped JDK 1.6.

@auhlig When do you anticipate the release of your V3 work and OpenStack4j version 3?

auhlig:
@gondor Creating, handling and merging in so many small PRs for the identity v3 work was good in terms of reviewing the code but took way longer than actually writing all of it. 
Anyway, I think you can start the countdown as we're about 4 PRs away from the release.

gondor:
@auhlig Thats exciting news!  You guys have done a fantastic job. Be proud when we release version 3 :)

@sachinwalia2k8 Come to think of it I don't think bumping up Jersey will hurt anything.  I forgot I created two separate connectors.  The default connector which is why this would address and then a special jersey-jdk16 openstack4j connector that can be used instead which is running on an older version.  Maybe I'm missing something else from this request.

gschukin:
I think that the JDK baseline for Jackson 2.7 is JDK 1.7.

sachinwalia2k8:
@gschukin - JDK 6 is supported until Jackson 2.7. Beginning 2.8 it'll be JDK 7 and onwards only.

PR #642, opened 2016-04-27
upgrade to java 7 and jackson 2.7.3

auhlig:
Upgrade to Java 7 as discussed in #496.

In terms of id v3 I think that is the last thing to include. After that you could trigger the release. 
Makes sense @gondor and @dhague ?


dhague:
looks good

gschukin:
@auhlig can we have Jackson upgrade to 2.7 for V3? Reference issue #563

auhlig:
@gschukin: Done.

gondor:
LGTM! :)

Issue #643, opened 2016-04-27
NPE if there is no Content-Length in headers for HEAD responce


gschukin:
Swift may not return Content-Length header in responce for zero size objects for HEAD request.

A check is needed in ParseObjectFunction.apply(...) method whether Content-Length header is present in responce or not.

PR #644, opened 2016-04-27
NPE if there is no Content-Length in headers for HEAD responce #643

vinodborole:
Reference #643


vinodborole:
@gondor Do you think this is a valid defect to be merged?

gondor:
Its not that its invalid.  We are trying to release V3 this week.  If I merge it will only be in version 2.  We also have V3 locked down until release and won't be pulling in anymore code.

gondor:
Thank you @vinodborole for the vix

PR #659, opened 2016-05-11
Add managing OS::Heat::None

draoullig:
Hi,

Since Liberty, in the resource_registry we can use OS::Heat::None for disable or enables easily certain resources (http://docs.openstack.org/developer/heat/template_guide/openstack.html#OS::Heat::None).
I change the code, when this special entry is present, it doesn't try to load a content file and continue.
I made some tests in my use case and it seems to be good.

Bests regards.

Frédéric.


gondor:
Thanks for the update!

PR #661, opened 2016-05-12
v3_collaboration

auhlig:
Hey @gondor, @dhague, 

I updated v3_collab and resolved conflicts. Ready to merge & release when you are :)

Looks good?


gondor:
:( Not sure if you guys are around.  All merged but tests are failing (19
tests).   Something went wrong, seeing this error :

org.openstack4j.api.exceptions.ConnectionException:
java.net.ProtocolException: HTTP method PATCH doesn't support output
at
sun.net.www.protocol.http.HttpURLConnection.getOutputStream(HttpURLConnection.java:1082)
at
org.glassfish.jersey.client.internal.HttpUrlConnector$4.getOutputStream(HttpUrlConnector.java:385)
at
org.glassfish.jersey.message.internal.CommittingOutputStream.commitStream(CommittingOutputStream.java:200)
at
org.glassfish.jersey.message.internal.CommittingOutputStream.commitStream(CommittingOutputStream.java:194)

auhlig:
@gondor: Not good :( Tests were successful on my machine. I'm on it.

dhague:
We traced this to a bug in Java 1.7 - we didn't see it initially as @auhlig and I both run JDK 1.8...

auhlig:
Also according to https://java.com/en/download/faq/java_7.xml Java 7 will no longer receive updates so it makes sense to upgrade to Java 8. Now would be a good time for that. What do you think? @gondor @dhague

dhague:
Given that Java 7 is now EOL for over a year (no public updates since April 2015, and none planned), I think it would be a good idea to standardize on Java 1.8 as a minimum for OpenStack4j 3.0.

gondor:
Ok, lets do it.  I'm totally ok with this.

PR #663, opened 2016-05-12
upgrade to java 1.8

auhlig:
@gondor, @dhague:  As discussed in #661 upgrading to jdk 8 should solve the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: java
 
 jdk:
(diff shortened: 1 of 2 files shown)


auhlig:
Tests are green for oraclejdk8 but openjdk8 seems to be unknown

auhlig:
@gondor This PR looks now much greener and good to merge. I removed 'openjdk8' as it is not available in travis.

PR #669, opened 2016-05-17
Missing tenant builder

vinodborole:
Tenant Builder was missing; probably it was lost during recent 3.0 release.


gondor:
Thanks @vinodborole

Issue #666, opened 2016-05-13
3.0.0 version cannot create V2 tenant


wumingcheng:
when I want to create a v2 tenant ,the "Builders.tenant()" was missing.

auhlig:
@wumingcheng Thanks for pointing that out.
I'll fix that and add some integration-tests next week.
As a workaround for the meantime you could use the KeystoneTenant.builder()  directly.

gondor:
New 3.0.1-SNAPSHOT has been pushed

auhlig:
Looking at the Builders I think it makes sense to also divide by Identity version. So it would be clear which Role, User, etc. is being built or that a Group is not available when using v2.
So I would create a BuilderV2  for identity v2 and a BuilderV3 for id v3 which extends the current Builder-implementation.
That would follow the pattern used in the OSClient, OSClientSession, .. .

Does this make sense @gondor, @dhague ?

gondor:
I like that idea, will make things cleaner and allow for expansion

dhague:
Sounds good. Technically this could introduce a breaking change in 3.0, but if we do it quickly then I think that's OK - it's not like we have a big installed base of OS4J 3.0 users yet :-)
My feeling is that if we get it out quickly as 3.0.2 (maybe even 3.1) then we should be fine.

PR #671, opened 2016-05-17
Group based Policy feature implementation

vinodborole:
OpenStack Group-Based Policy (GBP) framework, designed to offer a new set of abstraction that allows to manage OpenStack infrastructure through declarative policy abstractions. It provides declarative abstractions for achieving scalable intent-based infrastructure automation.

PR #675, opened 2016-05-18
Dividing Builder-API by Identity version

auhlig:
As discussed in #666 this introduces the BuildersV2 and BuildersV3 to make it less likely to get confused by the 2 different Identity versions. 

What do you think @gondor , @dhague ?


dhague:
As discussed with @auhlig via PM just now, this idea of using `Builders.<object>()` for everything except Identity, where we use `BuilderV2.<object>()` and `BuilderV3.<object>()` doesn't feel quite right.
A better approach would be to stay with `Builders.<object>()` which would always use the latest version of an API (and a point release of OS4J whenever this changes), together with `Builders.<service>().<object>()` when a specific version is required.
This would give, for example:
`Builders.user()` creating a V3 user (in the current 3.0 release of OS4J)
`Builders.identityV3().user()` creating a V3 user regardless of OS4J release (>3.0)
`Builders.identityV2().user()` creating a V2 user regardless of OS4J release (>3.0)

This same pattern can then be used to support different versions of the other OpenStack services, and has the advantage that users of OS4J v3.0 would not have a breaking change when using V3 user objects.

gondor:
@auhlig @dhague I agree with the ```Builders.<service>.<object>``` approach.  I was thinking about it when I was getting ready for work and have V2 and V3 builders felt odd to me and would impose lots of breaking changes in the future.  I like rooting it to the service based on the version of the service being used.    Well promote less breaking changes.

vinodborole:
Builders.<service>.<object> makes more sense

gondor:
@auhlig @dhague - I know this is a PR :) But wanted to introduce @vinodborole who made a comment above.  He's been a big contributor as well.  Do you mind also reviewing his latest PR which is targeted for V3 - PR #671  .  @vinodborole - Darren and Arno did the V3 work, infact built the whole 3.0.0 release.

vinodborole:
@gondor  Thanks for introducing me to the team; its a privilege working with you guyz. Let me know what you think about my PR;

wumingcheng:
@auhlig when does the new version publish, two weaks has gone.And don't forget to amend the bug of #679,

gondor:
Currently its done manually by me.  Working on integrating merged changes to deploy new snapshots via Travis CI.

auhlig:
Hey @gondor, @dhague, @vinodborole,

gondor:
LGTM - great job!

PR #681, opened 2016-05-23
Closing input stream, ref#622

vinodborole:
public static long copy(InputStream from,
        OutputStream to)
                 throws IOException
Copies all bytes from the input stream to the output stream. **Does not close or flush either stream**.

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/common/DLPayloadEntity.java
+++ b/core/src/main/java/org/openstack4j/openstack/common/DLPayloadEntity.java
@@ -42,8 +42,9 @@ public class DLPayloadEntity implements DLPayload {
     @Override
     public void writeToFile(File file) throws IOException {
         checkNotNull(file);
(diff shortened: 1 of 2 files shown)


vinodborole:
@auhlig Can you please review this request?

gschukin:
Why not to use try-with-resource since we're on 1.8 anyway?

vinodborole:
@gschukin thanks; that was a good suggestion. I have modified the PR with the change

gschukin:
Why not both streams?
I'm OOO so not able to comment properly :-(

vinodborole:
@gschukin thanks again, made the change.

gschukin:
So why not to stick to Outputstream interface? 

I'm preparing a big pr with findbugs and etc changes with all that small fixes. ;-)

vinodborole:
@gschukin that is really great! please let me know if i can be of any help in building that PR

Issue #573, opened 2016-02-14
Cannot create Flavors


starbender:
Looks like there are two properties being sent to Nova to create a Flavor which it doesn't like. 
OS-FLV-DISABLED:disabled and public

Getting the following error: 
`Additional properties are not allowed (u'OS-FLV-DISABLED:disabled', u'public' were unexpected), status=400, status-code=BAD_REQUEST}
`
Using the following code.
```
Flavor flavor = os.compute().flavors()
                .create("name",512,1,10,0,0,0f,true);

```
Connecting to a v2.0 endpoint.

I can't tell if this is a version problem of API I'm connecting to or not.

lakersan2015:
version 2.0.9 / 2.10

Additional properties are not allowed (u'OS-FLV-DISABLED:disabled', u'public' were unexpected)

GoForce5500:
I've got the same problem in 2.0.9, could anyone tell me if openstack4j supports OpenStack Liberty?

elikraja:
Same issue using 2.11 version. I'm using Openstack Liberty.

vinodborole:
@elikraja  Are you getting the OSClient with the perspective as ADMIN?

           OSFactory.builder().endpoint(authURL).credentials(username, password).tenantId(tenantID).useNonStrictSSLClient(true).perspective(Facing.ADMIN).authenticate();

elikraja:
Hi @vinodborole 
I wasn't, but nothing changed using the perspective(Facing.ADMIN). I got the same error

vinodborole:
@elikraja Did you check if you can create the flavor using openstack CLI with the same user?

elikraja:
Yep, first thing I tested was the creation from CLI.
I also tried the plugin using admin user. Nothing changes

vinodborole:
@elikraja  i am able to create it with o4j. My openstack version is Juno

My code:

     public static OSClient buildACIClient(){
        return OSFactory.builder().endpoint(authURL).credentials(username, password).tenantId(tenantID).useNonStrictSSLClient(true).perspective(Facing.ADMIN).authenticate();
    }

    private static void createFlavor(OSClient client) {
        Flavor flavor = client.compute().flavors().create("deleteme",512,1,10,0,0,1.0f,true);
        System.out.println(flavor.getId());
    }



INFO: 1 * Sending client request on thread main
1 > POST http://10.193.180.230:8774/v2/8be60cc30edc44978d24d51ce5f8d9da/flavors
[...]

PR #683, opened 2016-05-24
Removed parameters from NovaFlavor and respective classes to address #573

vinodborole:
@gondor @auhlig not sure if we want to modify existing classes or have a new version for this. Please review


menihillel:
Hi Vinod, Thanks for quick response. Just not sure if removal of the fields would be the right approach... those fields are still needed for read. I think we just nulling out the field and force Jersey to omit null fields might be a better approach.

vinodborole:
#686

PR #688, opened 2016-05-25
 	add SHELVED and SHELVED_OFFLOADED TO server STATUS

ustbxyls:
SHELVED and SHELVED_OFFLOADED  are supported in kilo


gondor:
LGTM, thanks

PR #690, opened 2016-05-26
best reading nested template

draoullig:
Hi,

If you use nested template and you have a mixed in the type property of a resource of relative path, absolute path, Http URL, Https URL, the loading of the nested templates failed because the path isn't correct.
I push a new method to calculate when the baseurl is needed or not to get the correct path to the file.

Cheers,

Frédéric


gondor:
LGTM, thank you

PR #693, opened 2016-05-30
Fix #673: Bring back Java 7 support

olivergondza:
- Project depends on 1.7
- Controllers affected by [JDK-7157360](https://bugs.openjdk.java.net/browse/JDK-7157360) depends on 1.8
- `distribution` module depends on `openstack4j-resteasy`

PR #704, opened 2016-06-01
Fixup 87237fa61a3982f41a12742f5ad15e7a6dc0efc5

olivergondza:
This fixes a mistake made in https://github.com/ContainX/openstack4j/pull/693.

### DIFF
--- a/core-integration-test/src/test/groovy/org/openstack4j/api/AbstractSpec.groovy
+++ b/core-integration-test/src/test/groovy/org/openstack4j/api/AbstractSpec.groovy
@@ -18,7 +18,7 @@ import co.freeside.betamax.tape.yaml.TapePropertyUtils
 abstract class AbstractSpec extends Specification {
 
     def static String MODULEROOT = Paths.get("").toAbsolutePath().getParent().toString()
-    def static String TAPEROOT = "/" + MODULEROOT + "src/test/resources/tapes/"
+    def static String TAPEROOT = MODULEROOT + "/src/test/resources/tapes/"
     def static Config CONFIG_PROXY_BETAMAX = Config.newConfig().withProxy(ProxyHost.of("http://localhost", 5555))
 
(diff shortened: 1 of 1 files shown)


olivergondza:
The integration tests are still failing, though.

core-integration-test/src/test/groovy/org/openstack4j/api/AbstractSpec.groovy
```diff
@@ -18,7 +18,7 @@ import co.freeside.betamax.tape.yaml.TapePropertyUtils
 abstract class AbstractSpec extends Specification {
 
     def static String MODULEROOT = Paths.get("").toAbsolutePath().getParent().toString()
-    def static String TAPEROOT = "/" + MODULEROOT + "src/test/resources/tapes/"
+    def static String TAPEROOT = MODULEROOT + "/" + "src/test/resources/tapes/"
```
auhlig: @olivergondza I think it should be TAPEROOT = "/" + MODULEROOT + "/src/test/resources/tapes/"

With JDK 1.7 that reduces the amount of errors in the integration-test module to 3 NPE's caused by UnknownHostExceptions. Very strange. I'm currently looking into it.

olivergondza: That is strange, the absolute path should start with '/' so additional 
should be ignored and it will blow up on Windows completely. Can we 
agree on `MODULEROOT + "/src/test/resources/tapes/"`?

auhlig: Yes, makes sense.

The remaining 3 issues, which prevent os4j from going back to green, are still unclear to me. 
Maybe it's something in Betamax. Need to dig in this again tomorrow - unless you're faster.

olivergondza:
Rebased, with what we agreed on. This is ready to be merged.

PR #705, opened 2016-06-02
Introduce base for micro versions

gonzolino:
Many OpenStack Services, like nova and manila, use micro-versions for their APIs.

This PR introduces a new base class for micro-versioned services, which will send an HTTP header specifying the desired micro-version with each API request.

Omitting this HTTP header causes the API to fall back to its minimum version (e.g. 2.1 for nova, 2.0 for manila), which prevents OpenStack4j to use newer API features.

This PR also includes a patch which enables micro-versions for manila. The latest version supported in this patch is 2.6, which equals the latest micro-version supported in Liberty. I will continue to work on supporting newer API versions (the latest micro-version for manila is 2.15 at the moment).

Please note that I consider this PR as a basis for discussion and that I am open for suggestions on how to improve this design. For example, there is no way for the user to specify which micro-version to use.

### DIFF
--- a/core-test/src/main/java/org/openstack4j/openstack/internal/MicroVersionedServiceTest.java
+++ b/core-test/src/main/java/org/openstack4j/openstack/internal/MicroVersionedServiceTest.java
@@ -1,0 +1,96 @@
+package org.openstack4j.openstack.internal;
+
+import org.openstack4j.api.AbstractTest;
+import org.openstack4j.api.types.ServiceType;
+import org.openstack4j.core.transport.HttpMethod;
+import org.openstack4j.core.transport.HttpRequest;
+import org.openstack4j.openstack.internal.MicroVersion;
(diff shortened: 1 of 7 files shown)


gondor:
@ContainX/openstack4j team - please review.  I think this is a pretty good approach to offer more reliable extensions.

auhlig:
@gonzolino Could you at some tests? Besides that: LGTM

gonzolino:
@auhlig that's a great suggestion.
I will look into testing next week!

vinodborole:
LGTM

gonzolino:
Newest commit contains tests for MicroVersion and MicroVersionedOpenStackService.

gondor:
Great, LGTM.  Going to merge and you will see build failures.  Its due to a recent change to support JDK7 again but JDK8 is failing on test.  I'll still deploy a new snapshot

Issue #707, opened 2016-06-03
failing build


auhlig:
With help of @dhague we figured out what's currently causing the failing build after #693.
Most of the problems got fixed after correcting the path to the tapes-folder by #704, the remaining 3 failures where caused by the different way parameters are appended to a uri using jdk 1.7 and 1.8 .
Apparently Betamax just compares the strings. -.-
For example:
```http://devstack.openstack.stack:5000/v3/users?name=foobar&domain_id=default``` vs ```http://devstack.openstack.stack:5000/v3/users?domain_id=default&name=foobar``` .
If there's  no exact match we get the NPE as seen in the build logs, because betamax is unable to find a response to play back .

After correcting that in the tapes, and still skipping [JDK-7157360](https://bugs.openjdk.java.net/browse/JDK-7157360)-affected controllers using ```mvn clean verify -pl '!org.pacesys.openstack4j.connectors:openstack4j-http-connector,!org.pacesys.openstack4j.connectors:openstack4j-jersey2,!org.pacesys:it-jersey2' ```we get OS4j back to green using JDK 1.7 .

Just to be clear: It would still fail using JDK 1.8 as it is sorting parameters differently. 

So if we want to get the integration-tests back in both java versions we could either try to fix that in betamax or try to somehow fix the ordering of parameters in os4j itself.

For now I could follow up with a PR which works for 1.7 or 1.8 . 
What do you guys think makes most sense here?

@gondor, @dhague, @vinodborole 

/cc @olivergondza

vinodborole:
as with reducing the minimum requirement for os4j to 1.7; for now a PR for 1.7 LGTM

gondor:
Hey team, Personally I think its fair to the community to support Current JDK - 1 version.  After hearing a lot of feedback I think thats fair.  When JDK9 is officially release we could then drop JDK7.  It'd be nice to set the tone with the community and how we decide on JDK versions.

We're very close as you mentioned to workaround the JDK7 bug and offer support for both JDK's.  

As far as betamax vs sorting I almost feel the patch should be in betamax?  I could fork it on ContainX so we can control the relationship between OS4J and Betamax.  It seems like parameter ordering may be a larger refactor but maybe I'm wrong.  

How do you feel about this?

olivergondza:
@gondor, I am afraid that forking dependencies for rather trivial fixes might be a slippery slope as you and up maintaining outdated forks with modifications you depend on (making it harder to switch to the upstream again).

[...]

PR #708, opened 2016-06-04
fix failing integration-tests

auhlig:
Hi,
As mentioned in #707,
this should do the trick to fix the integration-test failures at least for JDK 1.7 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 language: java
 
 jdk:
-  - oraclejdk8
+  - oraclejdk7
+  - openjdk7
 
 notifications:
   email: false
 
-script: "mvn verify"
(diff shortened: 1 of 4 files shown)


auhlig:
Looks the script in the travis.yml, which should skip test for connectors affected by the jdk1.7-bug didn't work - although it works locally.

olivergondza:
It seems to skip the `script` instruction altogether and always run `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V`.

auhlig:
Yes. I'm not sure why the instruction works locally but is ignored by travis.

gondor:
Let me try updating the master .travis with these changes.  I have a feeling travis is looking at masters.

olivergondza:
This seems to be the cause: https://docs.travis-ci.com/user/languages/java#Dependency-Management-with-Default-install

auhlig:
closing this, because of #710

PR #710, opened 2016-06-06
Fix #707: Fix the build skipping the Travis' install step

### LINKED ISSUE #707: failing build

olivergondza:


olivergondza:
Now I am happy with what it does and I managed to run the CI on both JDK 7 and 8. I am not really fluent with `.travis.yml` so please feel free to simplify. @auhlig, it seems the fix for JDK7 (reordering arguments) breaks the build for JDK8.

auhlig:
@olivergondza: Awesome! Thanks for fixing.

Yes. That was to be expected. 
But it looks like the JDK1.8 version is not able to find the tapes at all. Must be caused by the recent changes to lines https://github.com/ContainX/openstack4j/blob/master/core-integration-test/src/test/groovy/org/openstack4j/api/AbstractSpec.groovy#L20-L21.

But maybe JDK1.7 is good for now and will give us some time to address the root cause in betamax itself.

PR #713, opened 2016-06-08
travis configuration for only jdk1.7

auhlig:
For now taking @olivergondza's working solution to get os4j back to green on jdk1.7 . 

I'll look into getting the integration tests with jdk1.8 back again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,22 +4,13 @@ jdk:
   - oraclejdk7
   - openjdk7
 
-env:
-  global:
-    - SKIP_JDK8_MODULES='-pl !org.pacesys.openstack4j.connectors:openstack4j-http-connector,!org.pacesys.openstack4j.connectors:openstack4j-jersey2,!org.pacesys:it-jersey2'
-
-matrix:
-  include:
-    - jdk: "oraclejdk8"
-      env: SKIP_JDK8_MODULES=""
-
 notifications:
   email: false
 
+
 install: /bin/true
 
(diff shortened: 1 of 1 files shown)


olivergondza:
I am not sure I follow the change. Is this a work in progress?

auhlig:
@olivergondza: As https://github.com/ContainX/openstack4j/pull/710/commits/d08dc41fe5d4f7db8aa8e41cc3edfbe42e0bfa1c didn't work and I require some time to get the integration tests working with jdk1.8 I though we could at least for jdk1.7 get the build status back to green for now. Makes sense?

olivergondza:
oh, ok.

auhlig:
Maybe we can put this PR on hold for now, because we might get os4j on jdk1.8 back to green quicker than I initially thought: PR to address our query paramete ordering issue to Betamax is submitted: https://github.com/betamaxteam/betamax/pull/207 .
Let's see what comes out of that.
/cc @ContainX/openstack4j

gondor:
@auhlig I noticed betamax merged your change.  How often to they update their snapshots or releases?  Can we refer to a snapshot if needed?

auhlig:
@gondor I saw that too. I think

auhlig:
@gondor: Sorry for the delay. I ran into some problems when trying to integrate the above mentioned fix to betamax in os4j. We had to figure out the root cause and submit another PR to betamax just today.

Good news: I already verified that this fixes os4j's jdk1.8 build status. We can integrate that as soon as the PR is merged and a new version of betamax is deployed.

Issue #684, opened 2016-05-24
I would like to add Senlin module to the project.


L-I-O-N:
#680 As openstack mitaka version add a new core project which is called Senlin. 
I have finished writing the code and submitted PR [https://github.com/ContainX/openstack4j/pull/680]

auhlig:
@L-I-O-N I guess after the PRs were merged this issue can be closed?

L-I-O-N:
Yes

PR #714, opened 2016-06-08
Add senlin model

L-I-O-N:
As openstack mitaka version add a new core project which is called Senlin. According to the basic framework, I Add a new module to implements Senlin RESRful API.And I hava tested all the basic functions of this module. #684


auhlig:
Why are you introducing the ActionID ? Shouldn't the actions id be an attribute in the, already existing, Action?

L-I-O-N:
This is because the return of the action and the acquisition of the action format is different.
{
    "action": {
        "action": "CLUSTER_SCALE_IN",
        "cause": "RPC Request",
        "created_at": "2016-06-15T02:42:38.000000",
        ........
    }
}
and
{
    "action": "40a436b1-28d1-4de6-b2c3-0a34f478e2c9"
}

L-I-O-N:
@gondor Any question? This matter has been on hold for almost a month.

vinodborole:
@L-I-O-N Apologies for the delay and thank you for your patience.

@auhlig @gondor  LGTM

L-I-O-N:
Thanks for your support. @vinodborole. Our projects have used this SDK,so there is no obvious problems on function.

core/src/main/java/org/openstack4j/openstack/senlin/domain/SenlinAction.java
```diff
@@ -0,0 +1,199 @@
...
+		return target;
+	}
+
+	@Override
+	public Integer getTimeout() {
+		return timeout;
+	}
+
+	@Override
+	public String toString() {
```
auhlig: Why not use something like 
```
public String toString() {
  return Objects.toStringHelper(this).omitNullValues()
                .add("action", action)
                .add("cause", cause)
                // ...
                .toString();
    }
```

auhlig:
I hesitate to follow your style. 
Just one example: Looking at [SenlinClusterCreate.java](https://github.com/L-I-O-N/openstack4j/blob/1624fcfca167851c6c67d3ba2469357588504815/core/src/main/java/org/openstack4j/openstack/senlin/domain/SenlinClusterCreate.java).
Comparing your implementation to other parts of os4j I'd suggest to create a class with @JsonRootName("cluster") and attributes like name, max_size,... from the response instead of creating a cluster-map and adding all those there.

PR #715, opened 2016-06-12
Implementation for instance action logs

oxisto:
This is an implementation for the nova extension _os-instance-actions_. It follows the same code paradigm as the implementation for _os-interface_


core/src/main/java/org/openstack4j/openstack/compute/internal/ext/InstanceActionsServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+	public List<? extends InstanceAction> list(String serverId) {
+		checkNotNull(serverId, "serverId");
+		return get(NovaInstanceActions.class, uri("/servers/%s/os-instance-actions", serverId)).execute().getList();
+	}
+
+	@Override
+	public InstanceAction get(String serverId, String requestId) {
+		checkNotNull(serverId, "serverId");
+		checkNotNull(requestId, "requestId");
+		return get(NovaInstanceAction.class, uri("/servers/%s/os-instance-actions//%s", serverId, requestId)).execute();
```
auhlig: Isn't there a '/' in ```/servers/%s/os-instance-actions//%s``` too much?

core/src/main/java/org/openstack4j/openstack/compute/internal/ext/InstanceActionsServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    public List<? extends InstanceAction> list(String serverId) {
+        checkNotNull(serverId, "serverId");
+        return get(NovaInstanceActions.class, uri("/servers/%s/os-instance-actions", serverId)).execute().getList();
+    }
+
+    @Override
+    public InstanceAction get(String serverId, String requestId) {
+        checkNotNull(serverId, "serverId");
+        checkNotNull(requestId, "requestId");
+        return get(NovaInstanceAction.class, uri("/servers/%s/os-instance-actions//%s", serverId, requestId)).execute();
```
auhlig: Isn't there a '/' in /servers/%s/os-instance-actions//%s too much?

auhlig:
Hi @oxisto,
Could you add some short tests?

core/src/main/java/org/openstack4j/openstack/compute/internal/ext/InstanceActionsServiceImpl.java
oxisto: Yes you are correct. I'll fix that and try to add a test case tomorrow.

oxisto:
Is there a specific test case which i can build upon? As far as i can see there is no tests for the other compute extenions APIs, such as os-interfaces, etc. - which makes building the test case from scratch a bit difficult for me.

auhlig:
@oxisto If you're referring to the integration-test you're right. That would initially be a little bit more work. 

But as this PR is pending for so long it would be also okay to add some mock-tests in https://github.com/ContainX/openstack4j/tree/master/core-test/src/main/java/org/openstack4j/api/compute .

oxisto:
Seems to work now but somehow I ended up committing a file which I didn't want to. I'll try to undo the mess, if I cannot managed to do that, i'll create a new PR.

oxisto:
Ok this totally destroyed the PR, sorry. I'll close this one and create a new one.

PR #720, opened 2016-06-15
Fix URL for manila back-end storage pools

### LINKED ISSUE #719: os.share().schedulerStats().pools() always returns empty list
Listing manila back-end storage pools does not work.
The result of pools() and poolDetails() is always an empty list.

Debugging of the call unveiled that the URL path of the call is wrong.
Currently it is "/pools" but it should be "/scheduler-stats/pools".

IMHO This issue should be fixed in both the master and the maintenance branches.

gonzolino:
This patch resolves #719


auhlig:
LGTM

Issue #701, opened 2016-06-01
Fully support v2 of Telemetry API


llarsson:
While trying to create an obvious fix to #700, I stumbled upon further errors. It seems as if the support in OpenStack4j for [Telemetry API v2](http://developer.openstack.org/api-ref-telemetry-v2.html) is incomplete. Aside from the previously mentioned issue, the [data model](http://developer.openstack.org/api-ref-telemetry-v2.html#showSample) has changed rather significantly, referring e.g. to "meter" as opposed to "counter name".

vinodborole:
@gondor do we want to update the existing data model with modified properties?

gondor:
@vinodborole Yes, telemetry was probably modelled around the Havana release timeframe.  Many versions back.  As a rule we've tried to maintain support for 2-3 versions back max.

sjpatil:
I will soon send PR for this

PR #728, opened 2016-06-24
Added implementation for ceilometer sample

sjpatil:
Earlier there was support for samples listed from meter service only " /v2/meters/​{meter_name}​".
Added implementation for sample service and APIs to support sample operation related to it "/v2/samples" and "/v2/samples/{sample_id}"
There is discrepancy in responses of APIs mentioned in openstack API docs for meter service samples as they are shown same as sample service samples but actually they are different.
This will support in part for issue #701


vinodborole:
@sjpatil 
please fix the errors

vinodborole:
@sjpatil  Fix your test cases please

sjpatil:
@auhlig  My build is failing in jersey2 connector integration tests , can you help here?

auhlig:
Short answer: It fails only with jdk1.8. That's not related to your work. There's still an issue with jdk1.8 and the integration-tests.

Longer answer: The current problem is the integration of the recent version of betamax. After being distracted from that I'm now looking into solving that. So far betamax is not easy to debug -.-

vinodborole:
LGTM @gondor and @auhlig  can you guyz review?

auhlig:
LGTM

gondor:
LGTM, thanks for being patient

PR #729, opened 2016-06-27
Fix #734: add extra field on project

### LINKED ISSUE #734: Support extra field in project API.
Currently openstack's identityV3 api supports extra field on project and user.
The extra field can be sent to keystone with as top level json element and keystone send back the extra field as top level json element.

* For example, when calling project update API like followings.
```json
{
    "project": {
        "is_domain": true,
        "description": "My updated project",
        "links": {
            "self": "http://localhost:5000/v3/projects/93ebbcc35335488b96ff9cd7d18cbb2e"
        },
        "enabled": true,
        "id": "93ebbcc35335488b96ff9cd7d18cbb2e",
        "parent_id": null,
        "domain_id": "default",
        "name": "myUpdatedProject"
        "my_extra_key": "wow" <== HERE
    }
}
```

Keystone sends back extra fields as the top level json element. 
```json
{
    "project": {
        "is_domain": true,
        "description": "My updated project",
        "links": {
            "self": "http://localhost:5000/v3/projects/93ebbcc35335488b96ff9cd7d18cbb2e"
        },
        "is_domain": true
        "enabled": true,
        "id": "93ebbcc35335488b96ff9cd7d18cbb2e",
        "parent_id": null,
        "domain_id": "default",
        "name": "myUpdatedProject",
        "my_extra_key": "wow"   <== HERE
    }
}
```

Currently openstack4j doesn't support extra field yet. However, the extra field is tje very important feature when developers make a system based on keystone and want to put some meta data on project.

junoyoon:
This includes the fix for #734 "add extra field on project

To support extra field, I have modified org.openstack4j.openstack.identity.v3.domain.KeystoneProject like followings.

* Put ```
    private Map<String, String> extra = new HashMap<String, String>();``` in the KeystoneProject
* Create getExtra / setExtra with ```@JsonAnyGetter``` and ```@JsonAnySetter``` which are called when there are no explicitly mapped field.
* Ignore __is_domain__ field in setExtra which is automatically calculated field from keystone. Currently openstack4j just ignores this field. However with this patch if we don't ignore this in setExtra, it will be re-sent to the keystone during project update call and keystone will keep this in DB's extra fields.
* Put @JsonIgnore on getLinks(), because if it's re-sent to keystone while project update, it will be interpreted as extra field just like __is_domain__.
* Put ```@JsonIgnoreProperties(value = "extra" , ignoreUnknown = true)``` on  KeystoneProject class. If we don't provide this on KeystoneProject class, ```private Map<String, String> extra = new HashMap<String, String>(); ``` will be translated to ``` "extra": {  ....  }```. This is not working in IdentityV3. Puting ```@JsonIgnore``` on ```private Map<String, String> extra = new HashMap<String, String>(); ```  doesn't not working for @JsonAnyGetter and @JsonAnySetter as well. I found that this is the only way not to send extra field to keystone.

You can not find the exact blueprint for this in keystone but I can find relevant [the blueprint](https://blueprints.launchpad.net/horizon/+spec/support-extra-prop-for-project-and-user) in horizon.

Because it's only visible when the additional field is input, there is not much examples in OpenStack IdentityV3 spec. 

You can fine the some evidence in the following. __is_domain__ field.
http://developer.openstack.org/api-ref-identity-v3.html#updateProject

### DIFF
--- a/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneProjectServiceTests.java
+++ b/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneProjectServiceTests.java
@@ -22,7 +22,10 @@ public class KeystoneProjectServiceTests extends AbstractTest {
     private static final String PROJECT_DOMAIN_ID = "7a71863c2d1d4444b3e6c2cd36955e1e";
     private static final String PROJECT_DESCRIPTION = "Project used for CRUD tests";
     private static final String PROJECT_DESCRIPTION_UPDATE = "An updated project used for CRUD tests";
-	
+    private static final String PROJECT_EXTRA_KEY_1 = "extra_key1";
+    private static final String PROJECT_EXTRA_VALUE_1 = "value1";
+    private static final String PROJECT_EXTRA_KEY_2 = "extra_key2";
+    private static final String PROJECT_EXTRA_VALUE_2 = "value2";
     private String PROJECT_ID;
 
     @Override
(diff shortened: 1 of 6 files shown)


auhlig:
Hi @junoyoon,
Could you paste a link to the documentation of the feature above in the idv3 api?

junoyoon:
@auhlig  Added.

auhlig:
@junoyoon Thanks. I initially was aware of the is_domain but not of the extra field.

LGTM

vinodborole:
LGTM

gondor:
LGTM

PR #730, opened 2016-06-27
Fix #733 : add role assignment list api

### LINKED ISSUE #733: Support role assingment list API
Current openstack4j API doesn't support role assignment list APIwhich let the client know who has which permission on a project. 

Without this API, we should iterate all roles and users for a project to determine who have which roles in which project or domain.

* http://developer.openstack.org/api-ref-identity-v3.html#listRoleAssignments

To avoid too much consequent calls to keystone, the API should return retrieved userIDs, groupIDs, roleIDs, projectIDs by getting ID not resolving each element again.

junoyoon:
Fix for #733


vinodborole:
LGTM

auhlig:
Nice. LGTM.

gondor:
LGTM

PR #738, opened 2016-07-06
Fix integrationtests

auhlig:
*Finally* this fixes the bug in the it-module caused by different sorting of query paramteres in jdk1.7 and jdk1.8 by
- upgrading betamax to latest snapshot version
- use method, path, query params to match recordings in the tapes 

other:
- slightly updated the readme in integration-test module
- re-enable logging by adding missing dependency

@ContainX/openstack4j

### DIFF
--- a/core-integration-test/README.md
+++ b/core-integration-test/README.md
@@ -1,16 +1,23 @@
 openstack4j-integrationtest
 ===========================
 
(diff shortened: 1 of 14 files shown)


gondor:
YES!! :) Awesome @auhlig .   LGTM

auhlig:
Thanks :)
Merging now..

vinodborole:
@auhlig This is great stuff!

PR #739, opened 2016-07-06
API provider microversion support

gonzolino:
This PR adds support to the APIProvider to create and maintain services in certain micro versions. This allows users to specify which micro version they want to use for their service.

It also contains the necessary code to use this with manila. SO users can specify their desired micro version like this:

``` java
ShareService shareService = os.share(new MicroVersion(2, 6));
```

### DIFF
--- a/core/src/main/java/org/openstack4j/api/APIProvider.java
+++ b/core/src/main/java/org/openstack4j/api/APIProvider.java
@@ -1,5 +1,8 @@
 package org.openstack4j.api;
 
+import org.openstack4j.common.MicroVersionedRestService;
+import org.openstack4j.openstack.internal.MicroVersion;
+
 /**
  * To keep our dependencies simple in the current Openstack4J, we utilize ServiceLoader to load a provider who is responsible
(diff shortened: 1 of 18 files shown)


gondor:
@gonzolino Thank you!   Does change any behaviour to existing core services like Identity, Nova, etc?   Would you also be able to provide me with a quick right up on usage so we can document this on OpenStack4j?  Or even better your more than welcome to fork the openstack4j.com repo, add the change and submit a PR.

gonzolino:
@gondor At the moment this only affects manila. But it would be possible to add support for micro versions to nova using the mechanisms implemented here and in PR #705. I will have a look at the documentation and explain the usage of this feature there.

@auhlig is currently taking a look at my implementation to see if the creation of Service objects in `DefaultAPIProvider` can be done in a better way.

Currently objects are created with the default micro version (via `newInstance()`) and the desired micro version is set afterwards using a setter. In my opinion it would be better to to use the constructor which accepts the micro version as a parameter (`getConstructor(MicroVersion.class).newInstance(version)`) and remove the setter, but I could't find a way to do it.

So before merging we should wait if @auhlig comes up with a better way to implement this.

PR #743, opened 2016-07-09
Adding support for Network Service Policy APIs in Group Based Policy

sumit74:


gondor:
LGTM, thank you for adding tests.

Issue #585, opened 2016-02-29
any plan to support lbaas v2.0?


kkxue:
Hello,openstack4j have supported lbaas v1.0,but it's DEPRECATED.
refs: http://developer.openstack.org/api-ref-networking-v2-ext.html
so,do you have a plan to support lbaas v2.0?

emjburns:
Any update on this?

PR #750, opened 2016-07-13
proposed apis for lbaas v2 support

emjburns:
I'd like to add support for v2 lbaas (v1 is currently depreciated). Issue is: https://github.com/ContainX/openstack4j/issues/585 

Here is my proposition for high level APIs. I've named the overall service LbaasV2Service because in lbaas v2 they replace "vip" with "loadbalancer". I think that the service responsible for all load balancer ("/loadbalancer/...") api calls should carry the loadbalancer name. Because of that, I've given the overall service the 'lbaas' name.

Lbaas v2 apis for reference: http://developer.openstack.org/api-ref-networking-v2-ext.html#lbaas-v2.0

@gondor please review when you get a chance. If you are open to me adding support for lbaas v2 I'll start implementing.

### DIFF
--- a/core/src/main/java/org/openstack4j/api/networking/NetworkingService.java
+++ b/core/src/main/java/org/openstack4j/api/networking/NetworkingService.java
@@ -1,6 +1,7 @@
 package org.openstack4j.api.networking;
 
 import org.openstack4j.api.networking.ext.FirewallAsService;
+import org.openstack4j.api.networking.ext.LbaasV2Service;
 import org.openstack4j.api.networking.ext.LoadBalancerService;
 import org.openstack4j.api.networking.ext.NetQuotaService;
 import org.openstack4j.common.RestService;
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/openstack4j/api/networking/NetworkingService.java
```diff
@@ -58,7 +59,12 @@
      * @return the LBaaS service
      */
     LoadBalancerService loadbalancers();
-    
+
```
vinodborole: I do not see any internal API implementations; is this all it has?

emjburns: @vinodborole yes I haven't implemented yet. Wanted to get the OK to add lbaas v2 support before I started.

emjburns: I'm closing this, will open another PR once I've finished implementation.

emjburns:
I'm closing this, will open another PR once I've finished implementation.

auhlig:
@emjburns Looked very promising so far :)
Could you also add tests while writing the the service implementation?

gondor:
@emjburns Looks great so far

emjburns:
@auhlig yep of course!

Issue #744, opened 2016-07-11
Identity v2 change password will result in user being disabled


boonseng3:
I'm using version 3.0.1-SNAPSHOT, when i use v2 Identity to change password, the password is changed but the enabled field becomes false.

I' have check the request sent, the enabled field is always set to false.
There is no need to sent the enabled field in the request at all.
`{"
"user" : {
"id" : "f041b9115103453998f4ee8aabbb2b85",
"password" : "password1",
"enabled" : false
}`

auhlig:
@boonseng3 You sure the user was enabled beforehand?

boonseng3:
@auhlig Yes, the user was enabled beforehand. 
The issue is that the json for changing password included the enabled field which defaulted to false.

auhlig:
I can confirm this.
Looking at the [relevant code](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/identity/v2/internal/UserServiceImpl.java#L72-76) I'm wondering where the enabled comes from.
Explicitly setting this to true doesn't help.
I continue tomorrow on this.

boonseng3:
Yup, setting it true doesn't help. Change password only requires id and password. Enabled is not required. I couldn't find out why it is included in the request while other fields are not included as the KeystoneUser has all the fields of a user.

auhlig:
@boonseng3 Should be fixed by #752.
You can check by getting the daily snapshot build or in the next stable release.

PR #752, opened 2016-07-14
Idv2 fix

auhlig:
- remove domainId attribute from identity v2 user (must have sneaked in during id v2/v3 work)
- enabled-attribute of v2 user becomes a primitive boolean to fix the root cause of #744. Thanks @dhague


gondor:
@auhlig LGTM :)

PR #755, opened 2016-07-20
fix endpoint for identity service when no default tenant/project is available.

jzhoucliqr:
For unscoped token, if user have no default project/tenant, when authenticate, no service catalog is returned. Then for identity service just directly return the endpoint, not try get from catalog.
 
Also make changes to identity endpoint when there is both v2 and v3 available. So even if the catalog returned only contains v2 api endpoint, v3 api will still work. (In case some deployment both v2 and v3 is used.)

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/identity/internal/DefaultEndpointURLResolver.java
+++ b/core/src/main/java/org/openstack4j/openstack/identity/internal/DefaultEndpointURLResolver.java
@@ -126,6 +126,8 @@ public class DefaultEndpointURLResolver implements EndpointURLResolver {
         if (token.getCatalog() == null) {
             if (ServiceType.IDENTITY.equals(p.type)) {
                 return token.getEndpoint();
+            } else {
+                return null;
             }
         }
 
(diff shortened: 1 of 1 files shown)


auhlig:
What do you think on this @jzhoucliqr ?

jzhoucliqr:
Hi auhlig,

Sorry for the late reply. 

Yeah, I agree with you. This is kinds of a hack I don't like. But unfortunately we have some customer openstack installations (more than one) have the misconfiguration of identity catalog, (only v3 enabled but catalog shown as v2), which is out of our control and currently not able to correct the configuration. 
I guess for now we'll might have to maintain our fork for this. 

For the unscoped token part, I think we do need a fix since no catalog is returned. 

Thanks.

auhlig:
@jzhoucliqr 
 I agree that https://github.com/ContainX/openstack4j/pull/755/commits/1ae0e833dda794044b326cedeee8e653916be5e5#diff-662af5097a9fc14062d96c047323bb10R148 is better done in your branch. The rest should be fine.

jzhoucliqr:
@auhlig 
Updated the PR to remove the v2/v3 part. Thanks.

auhlig:
Is the Identity V2 https://github.com/ContainX/openstack4j/pull/755/files#diff-662af5097a9fc14062d96c047323bb10R113 really affected? I don't think we can do unscoped authentication in v2.

jzhoucliqr:
@auhlig 
Yes we can do unscoped token in v2, at least tested in my setup works. The use case is to get all tenants for a user.

auhlig:
LGTM

PR #771, opened 2016-08-10
adding vipPortId to loadbalancer, adding container refs to listener

emjburns:
I was missing vipPortId in the loadbalancer object, and the tls and sni container refs in the listener object. In my openstack liberty you can only update the defaultTlsContainerRef, so that's how I've implemented here.

@vinodborole @gondor please take a look. Thanks!


vinodborole:
LGTM

PR #773, opened 2016-08-11
adding terminated_https to listener protocol

emjburns:
Missing terminated_https option for listener protocol. 

@vinodborole what have you found to be the most reliable source of documentation in general for OpenStack? Struggling to figure out which is most up to date/gets maintained.


vinodborole:
LGTM

vinodborole:
@emjburns  i have not found anything reliable than this
http://developer.openstack.org/api-ref.html

PR #777, opened 2016-08-16
add stack tag capabilities

drmaas:
* allow stack updates and creates to specify tags
* get stack returns tags
* filter stack list by query params

### DIFF
--- a/core/src/main/java/org/openstack4j/model/heat/builder/StackCreateBuilder.java
+++ b/core/src/main/java/org/openstack4j/model/heat/builder/StackCreateBuilder.java
@@ -1,9 +1,9 @@
 package org.openstack4j.model.heat.builder;
 
-import java.util.Map;
-
 import org.openstack4j.common.Buildable;
 import org.openstack4j.model.heat.StackCreate;
+
+import java.util.Map;
 
 /**
  * This interface describes a builder for {@link StackCreate} objects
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/openstack4j/model/heat/builder/StackCreateBuilder.java
```diff
@@ -76,5 +76,12 @@
     * @return StackCreateBuilder
     */
     StackCreateBuilder files(Map<String,String> files);
+
```
emjburns: Funny indentation here

core/src/main/java/org/openstack4j/openstack/heat/domain/HeatStack.java
```diff
@@ -47,6 +47,8 @@
 	private List<GenericLink> links;
 	@JsonProperty("updated_time")
 	private String updatedTime;
+	@JsonProperty("tags")
```
emjburns: you shouldn't need this if the variable name is the same as the json name

core/src/main/java/org/openstack4j/model/heat/builder/StackCreateBuilder.java
drmaas: will reformat whole file, it was a bit odd.

core/src/main/java/org/openstack4j/openstack/heat/domain/HeatStack.java
drmaas: Looks like it is convention to be explicit.

drmaas:
@gondor

core/src/main/java/org/openstack4j/openstack/heat/domain/HeatStack.java
emjburns: Good to know. I've seen varying things across the project.

drmaas: Meaning, explicit for this class. Not sure about others.

drmaas:
@vinodborole

vinodborole:
LGTM

PR #748, opened 2016-07-11
Telemtry Resource and capabilities API implementation

sjpatil:
Telemetry's resource and capabilities implementation added

Issue #154, opened 2014-12-05
Ceilometer - Support query criteria when retrieving samples


gondor:
Currently when we ask Ceilometer for samples for a given meter we retrieve everything.  This issue is to add support for advance query capabilities.

Brief example of new functionality:

``` java
List<? extends Sample> samples = os.telemetry().meters()
                                   .samples("image", SampleCriteria.create()
                                      .timestamp(Oper.GT, startDate)
                                      .timestamp(Oper.LTE, endDate)
                                      .resource(resourceId)
                                  );
```

accjiyun:
Why not post the sample to docs? I think it need more content.

PR #781, opened 2016-08-21
Add query capabilities to new ceilometer sample

simonklb:
The newly implemented ceilometer sample api (https://github.com/ContainX/openstack4j/pull/748) did not come with the query capabilities (https://github.com/ContainX/openstack4j/issues/154). This minor patch should fix that.


vinodborole:
LGTM

PR #789, opened 2016-08-29
Changed package Status to status

biogerm:
changed hypothetical misplaced Status file into correct package


biogerm:
It's minor without any code change.

gondor:
good catch! thanks

PR #793, opened 2016-08-31
provide method to show recursively deep resources on a stack

drmaas:


drmaas:
@gondor @vinodborole

vinodborole:
LGTM

Issue #797, opened 2016-09-06
UserService in Identity V3 lacked the "changePassword" method


zizhongwei:
UserService in Identity V3 lacked the "changePassword" method:
![image](https://cloud.githubusercontent.com/assets/12456254/18268788/e511e402-7457-11e6-9318-9d6291a8f39a.png)
"changePassword" method  in the OpenStack API：
![image](https://cloud.githubusercontent.com/assets/12456254/18268827/157f48e6-7458-11e6-96da-f6a3df808440.png)

auhlig:
I don't remember this was part of the initial id v3 api where you could change the user's password by using the update(User) method.
Do you want to submit this enhancement or should I take care during the week?

zizhongwei:
I want to submit this enhancement, 3ks.

auhlig:
#803 has been merged. Closing.

PR #803, opened 2016-09-13
Add missing method for user service

zizhongwei:
Add missing method "changePassword" for UserService in Identity V3 
Issue  #797


auhlig:
Nice. Thanks for your contribution @zizhongwei.

PR #807, opened 2016-09-13
Add Neutron agent-list

biogerm:
Neutron agent-list function was supported.
Fixed a missing symbol bug from Trove merge

This change is contributed by Yin Zhang, Magnus Broman, Qin An
(@biogerm)


biogerm:
@gondor @vinodborole @auhlig Here is the new PR. The old one was abandoned. Please take a look. Thank you!

core/src/main/java/org/openstack4j/openstack/networking/domain/NeutronAgent.java
```diff
@@ -0,0 +1,189 @@
...
+    private String id;
+
+    private static final Logger LOG = LoggerFactory.getLogger(NeutronAgent.class);
+
+    /**
+     * local help method to parse a string with format "yyyy-MM-dd HH:mm:ss" into a Date
+     * @param str - string to be parsed
+     * @return Date
+     */
+    private Date parseDate(String str) {
```
auhlig: Just an idea: Could use https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/internal/Parser.java instead of doing it here.
Makes sense?

auhlig:
LGTM

core/src/main/java/org/openstack4j/openstack/networking/domain/NeutronAgent.java
biogerm: Sounds like a good idea. I am now working on next PR so I will make a note and include it later.

biogerm:
Waiting for this to be merged because this commit fixes compiling issue. @gondor @vinodborole

auhlig:
@biogerm Could you address the comment and resolve the conflict?

biogerm:
@auhlig Ah. I was thinking to update source code in the next PR but since there's conflicts again... ok I will push a new PR for this.

biogerm:
@auhlig I looked into this and here is what I found out about. Neutron Agent has a different time stamp format. It's different than other API calls say, image-list. So that's why there's a different parser just for Agent list. Should I move this parser to Parser class or just keep it here? What do you say?

auhlig:
One place for that kind of methods might be better than multiple in order to avoid unnecessary code duplication. So I would move it to the Parser class. Do you agree?

biogerm:
I'm thinking the same. Here's the new commit.

auhlig:
Cool. Thanks for all that work @biogerm

auhlig:
Hey @biogerm, 
Seems like the recent merges, also of your work in the other PRs, created some conflicts. Could you resolve them and rebase and squash your commits? Sorry for the additional work. I promise to merge this PR as soon as this is done! :)

biogerm:
@auhlig no problem. it's quick :)

auhlig:
Thank. Merging now

Issue #818, opened 2016-09-20
NovaFlavor have a wrong get : Boolean isPublic()


zizhongwei:
NovaFlavor have a wrong get : Boolean isPublic()
when the NovaFlavor object convert to JSON ,the JSON text have a  unexpected  property  'public' ,like this : 
![image](https://cloud.githubusercontent.com/assets/12456254/18656595/086dcefc-7f25-11e6-96dd-331e8b676f68.png)

so the create API retunt BAD_REQUEST : 
![image](https://cloud.githubusercontent.com/assets/12456254/18656483/ffe6947c-7f23-11e6-8d31-6566af35377b.png)

zizhongwei:
I will fix this during the week

auhlig:
@zizhongwei Can we close this issue?

zizhongwei:
@auhlig yes,we should close this issue.

PR #820, opened 2016-09-20
fix NovaFlavor , add @JsonIgnore to Boolean isPublic()

zizhongwei:
fix NovaFlavor , add @JsonIgnore to Boolean isPublic()  
issues : #818

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/compute/domain/NovaFlavor.java
+++ b/core/src/main/java/org/openstack4j/openstack/compute/domain/NovaFlavor.java
@@ -141,6 +141,7 @@ public class NovaFlavor implements Flavor {
 	/**
 	 * {@inheritDoc}
 	 */
+	@JsonIgnore
 	@Override
 	public Boolean isPublic() {
 		return isPublic;


auhlig:
@zizhongwei Why ignoring the attribute? From what I can see here https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/compute/domain/NovaFlavor.java#L45 only `os-flavor-access:is_public`is being fetched and just `public`is ignored anyway, isn't it?

zizhongwei:
In class NovaFlavor , the get method named "isPublic()" returns a Boolean value. when the object of  this class was converted to JSON ,the JSON String will have a property 'public' ,But we also added the @JsonProperty("os-flavor-access:is_public") on the same property ,so the converted JSON String will have two property : "public" & "os-flavor-access:is_public" in JSON String, but the  property : "public"  is unexpected ;like this:
image

when use this  JSON String to call the create API,it will retunt BAD_REQUEST : 
https://cloud.githubusercontent.com/assets/12456254/18656483/ffe6947c-7f23-11e6-8d31-6566af35377b.png

so we need to  add @JsonIgnore on "Boolean isPublic()" to remove the unexpected property 'public' and remain the property "os-flavor-access:is_public"  in the  JSON String

auhlig:
Mh. Not quite sure where the `public` is coming from as the annotation cleary says `os-flavor-access:is_public` but your explanation sound reasonable.  What do you think @vinodborole?

vinodborole:
same thought @auhlig need more time to investigate on this, what do you think?

auhlig:
Any volunteers? ;)

biogerm:
We have seen same problem here.

biogerm:
LGTM

biogerm:
```
Sep 23, 2016 5:23:41 PM org.glassfish.jersey.filter.LoggingFilter log
INFO: 1 * Sending client request on thread main
1 > POST https://10.33.252.104:8774/v2.1/flavors
1 > Accept: application/json
1 > Content-Type: application/json
1 > User-Agent: OpenStack4j / OpenStack Client
1 > X-Auth-Token: gAAAAABX5UfQgQ0O8yVqgMQPY5E7M4fiWPpMv250h8rs0jspjjt7tgirdm3IRhnqbeM7Bn2AP3zmYOBI4prsit0PYS5ELOO4b1zBLLN_YiJPy2sXrgQ8o-9MUFCpXzcssXoSzWJzgQcIcRQrqrR6L1UaRztTkjqPH9UX0Yh68NDtKfHG4Tn5eyM
{
  "flavor" : {
    "name" : "testFlavor",
    "ram" : 4096,
    "vcpus" : 1,
    "disk" : 30,
    "swap" : 0,
    "public" : true,
    "OS-FLV-EXT-DATA:ephemeral" : 0,
    "rxtx_factor" : 1.0,
    "os-flavor-access:is_public" : true
  }
}

Sep 23, 2016 5:23:41 PM org.glassfish.jersey.filter.LoggingFilter log
INFO: 1 * Client response received on thread main
1 < 400
1 < Connection: close
1 < Content-Length: 343
1 < Content-Type: application/json; charset=UTF-8
1 < Date: Fri, 23 Sep 2016 15:23:41 GMT
1 < Vary: X-OpenStack-Nova-API-Version
1 < X-Compute-Request-Id: req-e9357d30-467f-40e2-9f1d-00695c1e421c
1 < X-Openstack-Nova-Api-Version: 2.1
{"badRequest": {"message": "Invalid input for field/attribute flavor. Value: {u'name': u'testFlavor', u'ram': 4096, u'vcpus': 1, u'swap': 0, u'os-flavor-access:is_public': True, u'rxtx_factor': 1.0, u'OS-FLV-EXT-DATA:ephemeral': 0, u'disk': 30, u'public': True}. Additional properties are not allowed (u'public' was unexpected)", "code": 400}}
```

vinodborole:
@biogerm thanks for volunteering :+1:  LGTM  @auhlig I think we can merge this, unless you have anything

auhlig:
Thanks for the work @biogerm .
I agree @vinodborole. Good to merge.

PR #826, opened 2016-09-21
Fix #825 - NullPointerException in HttpClient connector

### LINKED ISSUE #825: NullPointerException for HEAD request with HttpClient connector
Using the Apache HttpClient connector, a NullPointerException is thrown if the response contains no entity (which is the case for a HEAD request).

I encountered the issue with `client.identity().roles().checkProjectUserRole()` which is maped to a HEAD request (http://developer.openstack.org/api-ref/identity/v3/index.html?expanded=check-whether-user-has-role-assignment-on-project-detail,check-whether-group-has-role-assignment-on-project-detail#check-whether-group-has-role-assignment-on-project).

Then, when the HTTP response is mapped to an `ActionResponse` in https://github.com/ContainX/openstack4j/blob/3.0.2/core/src/main/java/org/openstack4j/core/transport/functions/ResponseToActionResponse.java#L31, the entity is `null` in https://github.com/ContainX/openstack4j/blob/3.0.2/connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpResponseImpl.java#L127, which causes the `NullPointerException`.

```
java.lang.NullPointerException
	at org.openstack4j.connectors.httpclient.HttpResponseImpl.readEntity(HttpResponseImpl.java:127) [openstack4j-httpclient-3.0.2.jar:?]
	at org.openstack4j.core.transport.functions.ResponseToActionResponse.apply(ResponseToActionResponse.java:31) [openstack4j-core-3.0.2.jar:?]
	at org.openstack4j.core.transport.functions.ResponseToActionResponse.apply(ResponseToActionResponse.java:22) [openstack4j-core-3.0.2.jar:?]
	at org.openstack4j.core.transport.HttpEntityHandler.handle(HttpEntityHandler.java:36) [openstack4j-core-3.0.2.jar:?]
	at org.openstack4j.connectors.httpclient.HttpResponseImpl.getEntity(HttpResponseImpl.java:65) [openstack4j-httpclient-3.0.2.jar:?]
	at org.openstack4j.openstack.internal.BaseOpenStackService$Invocation.execute(BaseOpenStackService.java:193) [openstack4j-core-3.0.2.jar:?]
	at org.openstack4j.openstack.internal.BaseOpenStackService$Invocation.execute(BaseOpenStackService.java:187) [openstack4j-core-3.0.2.jar:?]
	at org.openstack4j.openstack.identity.v3.internal.RoleServiceImpl.checkProjectUserRole(RoleServiceImpl.java:53) [openstack4j-core-3.0.2.jar:?]
```

olivierlemasle:
Fixes issue #825

### DIFF
--- a/connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpResponseImpl.java
+++ b/connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpResponseImpl.java
@@ -13,6 +13,8 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 public class HttpResponseImpl implements HttpResponse {
 
     private static final Logger LOG = LoggerFactory.getLogger(HttpResponseImpl.class);
@@ -123,8 +125,13 @@ public class HttpResponseImpl implements HttpResponse {
     @Override
     public <T> T readEntity(Class<T> typeToReadAs) {
         HttpEntity entity = response.getEntity();
+        if (entity == null) {
+            // Normal case if the response has no content, e.g. for a HEAD request
+            return null;
+        }
         try {
-            return ObjectMapperSingleton.getContext(typeToReadAs).reader(typeToReadAs).readValue(entity.getContent());
+            InputStream content = checkNotNull(entity.getContent(), "Entity content should not be null.");
+            return ObjectMapperSingleton.getContext(typeToReadAs).readerFor(typeToReadAs).readValue(content);
         } catch (Exception e) {
             LOG.error(e.getMessage(), e);
             throw new ClientResponseException(e.getMessage(), 0, e);


auhlig:
Hey @olivierlemasle,
I'm surprised the tests didn't reveal this earlier. 
Is only the httpclient affected? 

Thanks for fixing!

olivierlemasle:
>  Is only the httpclient affected? 

I've tested this request with Jersey2 (no issue) and with okhttp (I encountered an issue but it seems to be unrelated and I've not investigated yet).

auhlig:
Thanks for sharing the info.

auhlig:
Looks good to you @vinodborole ?

vinodborole:
@auhlig  added a comment

auhlig:
Agreed. null-check makes sense @vinodborole. Could you update the implementation @olivierlemasle? Thanks in advance.

olivierlemasle:
Do you mean _remove_ the null check, @auhlig ?
i.e. remove `|| entity.getContent() == null` ?

auhlig:
I think Vinod meant moving that logic to a separate function. Right, @vinodborole ?

olivierlemasle:
@auhlig @vinodborole 
Oh I thought it because `entity.getContent()` should never be null (unlike `entity`, which can be null, for example in the case of a HEAD request).

olivierlemasle:
@auhlig @vinodborole I've updated the commit.

connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpResponseImpl.java
```diff
@@ -123,8 +125,13 @@ public String header(String name) {
     @Override
     public <T> T readEntity(Class<T> typeToReadAs) {
         HttpEntity entity = response.getEntity();
+        if (entity == null) {
```
vinodborole: Why not use checkNotNull(entity) here, this will make it more compact and readable, what do you think?

olivierlemasle: Because `checkNotNull(entity)` would throw a `NullPointerException` if `entity == null`, which is exactly the issue #825. The goal of this pull request is to accept without error the case where `entity == null`, e.g. for HEAD requests.

For example, in Jersey2 connector, `readEntity` will return `null` for a HEAD request.

vinodborole:
LGTM, @auhlig your take?

auhlig:
Thanks for explaining @olivierlemasle .

LGTM too, Vinod.

PR #837, opened 2016-09-27
Update links: gondor -> ContainX

olivierlemasle:
Fixes some URLs.
And the waffle graph will work again!

[Throughput Graph](https://waffle.io/ContainX/openstack4j/metrics)


auhlig:
Yeah :) 
Thanks for fixing that.

PR #843, opened 2016-10-07
Murano API: Implementing actions API

nmakhotkin:


auhlig:
Besides the question above this looks very good. Many thanks for contributing @nmakhotkin

nmakhotkin:
Hi @auhlig! 

Sorry but I don't see any question from you.

auhlig:
@nmakhotkin Can you see the comment now?

What do you think on this @vinodborole ?

core-test/src/main/java/org/openstack4j/api/murano/v1/ServicesTests.java
```diff
@@ -74,6 +104,9 @@ public void testCreateOneServiceFromMap() throws IOException {
         ObjectMapper mapper = new ObjectMapper();
         Map<String, Object> serviceMap = mapper.readValue(json, Map.class);
 
+        // Need to remove internal service info from test json string.
+        serviceMap.remove("?");
```
nmakhotkin: Yes. Let me explain. The whole murano service structure json looks as the following: 

```
{
  "key1": "val1",
  "key2": "val2",
  ...
  "keyN": "valN",
  "?": {
  ..<service internal data like id, type, name etc.>..
  }
}
```

So I divide the application class model onto 2 parts: service internal info (which is all in "?" key) and all other app-related info (arbitrary data). Obviously, serviceMap here contains both data and service info but there is assertion only for data part. Without deletion value in "?" key this assertion won't work.

nmakhotkin:
Thanks, now I see. 

Btw, this PR is WIP for now till I submit action API implementation itself.

nmakhotkin:
@auhlig, @vinodborole, PR is ready for review now.

auhlig:
Nice work.

core-test/src/main/java/org/openstack4j/api/murano/v1/ServicesTests.java
auhlig: Makes sense. Thanks for explaining.

nmakhotkin:
Thanks @auhlig @vinodborole!

Issue #859, opened 2016-10-20
the server api does not support tag?


whaon:

whaon:
> Server tags (servers, tags)¶
> 
> Lists tags, creates, replaces or deletes one or more tags for a server, checks the existence of a tag for a server.
> 
> Available since version 2.26
> 
> Tags have the following restrictions:
> 
> Tag is a Unicode bytestring no longer than 60 characters.
> Tags are case sensitive.
> ‘/’ is not allowed to be in a tag name
> Comma is not allowed to be in a tag name in order to simplify requests that specify lists of tags
> All other characters are allowed to be in a tag name
> Each server can have up to 50 tags.

I get the content above from the openstack official website:[http://developer.openstack.org/api-ref/compute/](http://developer.openstack.org/api-ref/compute/)
but it seems that openstack4j doesn't support it.

auhlig:
Looks like it's indeed missing. Would you like to add this?

auhlig:
Same here. Can this be closed now?

Issue #860, opened 2016-10-20
the return values of getting details of keypair are incomplete


whaon:
```
{
    "keypair": {
        "fingerprint": "44:fe:29:6e:23:14:b9:53:5b:65:82:58:1c:fe:5a:c3",
        "name": "keypair-6638abdb-c4e8-407c-ba88-c8dd7cc3c4f1",
        "public_key": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC1HTrHCbb9NawNLSV8N6tSa8i637+EC2dA+lsdHHfQlT54t+N0nHhJPlKWDLhc579j87vp6RDFriFJ/smsTnDnf64O12z0kBaJpJPH2zXrBkZFK6q2rmxydURzX/z0yLSCP77SFJ0fdXWH2hMsAusflGyryHGX20n+mZK6mDrxVzGxEz228dwQ5G7Az5OoZDWygH2pqPvKjkifRw0jwUKf3BbkP0QvANACOk26cv16mNFpFJfI1N3OC5lUsZQtKGR01ptJoWijYKccqhkAKuo902tg/qup58J5kflNm7I61sy1mJon6SGqNUSfoQagqtBH6vd/tU1jnlwZ03uUroAL Generated-by-Nova\n",
        "user_id": "fake",
        "deleted": false,
        "created_at": "2014-05-07T12:06:13.681238",
        "updated_at": null,
        "deleted_at": null,
        "id": 1
    }
}
```
i fetch the content above from the os official website:[http://developer.openstack.org/api-ref/compute/?expanded=show-keypair-details-detail](http://developer.openstack.org/api-ref/compute/?expanded=show-keypair-details-detail)
but openstack4j lost the fields:user_id,deleted,created_at,updated_at,deleted_at and id.

auhlig:
Right. Maybe they were introduced after https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/compute/domain/NovaKeypair.java was written. 
Do you want to add the missing attributes?

auhlig:
Merged PR. Can we close this issue?

PR #861, opened 2016-10-21
add missing attributes in Keypair issue:#860

whaon:
add missing attributes in Keypair issue: #860


core/src/main/java/org/openstack4j/model/compute/Keypair.java
```diff
@@ -35,5 +37,41 @@
...
+	 * time deleted
+	 * @return
+	 */
+	Date getDeletedAt();
+	
+	/**
+	 * 
+	 * @return id of the keypair
+	 */
+	Integer getId();
```
auhlig: Looking at http://developer.openstack.org/api-ref/compute/?expanded=create-or-import-keypair-detail#create-or-import-keypair - What's the `id` of the keypair?

whaon: consider the below response content that I get from the url you given
```
{
    "keypair": {
        "fingerprint": "44:fe:29:6e:23:14:b9:53:5b:65:82:58:1c:fe:5a:c3",
        "name": "keypair-6638abdb-c4e8-407c-ba88-c8dd7cc3c4f1",
        "public_key": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC1HTrHCbb9NawNLSV8N6tSa8i637+EC2dA+lsdHHfQlT54t+N0nHhJPlKWDLhc579j87vp6RDFriFJ/smsTnDnf64O12z0kBaJpJPH2zXrBkZFK6q2rmxydURzX/z0yLSCP77SFJ0fdXWH2hMsAusflGyryHGX20n+mZK6mDrxVzGxEz228dwQ5G7Az5OoZDWygH2pqPvKjkifRw0jwUKf3BbkP0QvANACOk26cv16mNFpFJfI1N3OC5lUsZQtKGR01ptJoWijYKccqhkAKuo902tg/qup58J5kflNm7I61sy1mJon6SGqNUSfoQagqtBH6vd/tU1jnlwZ03uUroAL Generated-by-Nova\n",
        "user_id": "fake",
        "deleted": false,
        "created_at": "2014-05-07T12:06:13.681238",
        "updated_at": null,
        "deleted_at": null,
        "id": 1
    }
}
```
I think the id means the identity of the keypair

whaon:
I just merge the code that resolve the issue: #859 
I try my best to make it correct but it may still have problems.Our Openstack environment is lower than 2.26 while the api is available since version 2.26(please refer the api document),so I can't make a test for my code.
Please **test** my code and if it has any problems,just contact me for free,Thanks!

core/src/main/java/org/openstack4j/model/compute/Keypair.java
auhlig: I'm blind

auhlig:
I'd rather like to ask you to extend or add some unit tests.

auhlig:
Cool. Many thanks for the work @whaon, @tongh-qtl. Looks good from my side. Do you agree @vinodborole ?

vinodborole:
LGTM

auhlig:
@whaon Could you create a PR for https://github.com/ContainX/openstack4j.com and add documentation for the newly introduced methods? Many thanks in advance.

whaon:
it's ok,but maybe I will do it a bit later

auhlig:
Hey @whaon,
Right. It's just markdown and a little amount of html if you want to set some links.

Issue #390, opened 2015-05-27
Support for newest Service Versions


gondor:
Currently we always default to Compute V2 but as OpenStack progresses we should automatically select the newest endpoint version (for example V2.1 if it's deployed).

This issue will add support to always use the latest version.  Another issue will be created to allow for more granular control but that will be at a later cycle.

bsemp:
The patch for endpoint latest version selection works as expected but __should not be used for the moment__:

Endpoint latest version selection works:

```java
Server s = client.compute().servers().get("108b91cd-1adc-4799-b467-a7ec6fcd8ee7");
```
    2015/05/29 11:37:26:017 CEST [DEBUG] wire - http-outgoing-1 >> "GET /v2.1/b0b5ed7ae06049688349fe43737796d4/servers/108b91cd-1adc-4799-b467-a7ec6fcd8ee7 HTTP/1.1[\r][\n]"

```java
Volume v = client.blockStorage().volumes().get("8a9287b7-4f4d-4213-8d75-63470f19f27c");
```
    2015/05/29 11:42:36:959 CEST [DEBUG] headers - http-outgoing-1 >> GET /v2/b0b5ed7ae06049688349fe43737796d4/volumes/8a9287b7-4f4d-4213-8d75-63470f19f27c HTTP/1.1


__However, several bugs occurs related to the use of these new endpoints:__

 - Volume information are not complete :
```java
Volume v = client.blockStorage().volumes().get("8a9287b7-4f4d-4213-8d75-63470f19f27c");
System.out.println("Volume name=" + v.getName());
```
    2015/05/29 12:04:47:699 CEST [DEBUG] wire - http-outgoing-1 << "{"volume": {"status": "in-use", [...], "name": "test_vol", [...], "id": "8a9287b7-4f4d-4213-8d75-63470f19f27c", "size": 100}}"
    Volume name=null

 - I haven't tested other fields than volume.name nor other objects images/servers/snapshots so other similar bugs can be present.

 - Compute api 2.1 return HTTP error 500 with my deployment (ok, maybe it's a problem with my deployment, v2 works ok, v2.1 fails) but I ran into __connection leak problem again__...
```java
for (int i=0; i<=5; i++) {
    try {
        Server s = client.compute().servers().get("108b91cd-1adc-4799-b467-a7ec6fcd8ee7");
        System.out.println("Server=" + s);
    } catch (Exception e) {
        System.out.println("Exception occurs: " + e.getMessage());
    }
}
```

[...]

PR #862, opened 2016-10-22
Include the EndpointURLResolver in the services defined in the

bcornelis:
Include the EndpointURLResolver in the services defined in the
APIProvider. This allows for easy extending/overriding of the default
EndpointURLResolver logic.


auhlig:
Hi @bcornelis,
Thanks for contributing.
Is there an issue with the current way the EndpointResolver can be used/overridden? Could you share how your proposal would make it easier?
Also should that really live in apiprovider? What's your opinion @vinodborole, @gondor ?

bcornelis:
Dear @auhlig ,
the problem for me is that, as far as I can see, it's impossible to change the logic of the EndpointURLResolver. This one is instantiated in the OSClientSession class, and does not have any getters/setters. So how is it possible in the current situation, to provide a custom implementation?
As 'all' other services are defined in the Apis class, I thought it would be most application-compliant way to also use this logic to specify which implementation to use.

auhlig:
Currently it can be done the following way https://github.com/ContainX/openstack4j/issues/390

bcornelis:
Thanks for your explanation and sample, but this is not what I need. My usecase is that I need the identity V3, but the endpoint URL is not the same as the one returned in the tokens catalog. I Don't need the URL from the catalog, but the one specified in token.endpoint.

I think it's not possible to achieve this with the ServiceVersionResolver mentioned in the ticket #390, as for the V3 URL resolution, this resolver isn't even used. The only location where this ServiceVersionResolver is used, is to resolve the V2 (DefaultEndpointURLResolver::resolveV2). The resolveV3 is never used.

So in my case I've extended the DefaultEndpointURLResolver by always returning the tokens endpoint for the Identity service, and fallback to the default resolution (based on the catalog) for other services. As this class is instantiated like 'EndpointURLResolver epr = new DefaultEndpointURLResolver();' in the OSClientSession class, it seemed for me impossible to extend it.

What's your opinion on this?

auhlig:
Right. It's not the purpose of the endpoint resolver as, per default, OpenStack services get their respective endpoint from the token catalog. 
Just to make sure: Does defining a custom resolver
```
final ServiceVersionResolver resolver = new ServiceVersionResolver() {
    @Override
    public Service resolve(ServiceType type, SortedSet<? extends Service> services) {
        // resolver logic; possibly ext. default logic
        return endpoint;
    }
};
```
and overriding standard behavior with `.withConfig(Config.newConfig().withResolver(resolver))` while building the OSClient not work?

bcornelis:
This doesn't do the trick for me. Problem is really that the Openstack installation I'm connecting to has identity V2 exposed in the endpoints, but the V3 is available, but not listed in the endpoints. So I always get the V2 where the V3 is available. So if I choose V3 authentication, V2 is always chosen as this is the one included in the token.
Extending the ServiceVersionResolver won't do the trick, as this class is only used in the code for the V2 resolution, but I'm using V3.

vinodborole:
@bcornelis  "Problem is really that the Openstack installation I'm connecting to has identity V2 exposed in the endpoints, but the V3 is available, but not listed in the endpoints" - I think a better way to deal with this is you fix your openstack rather than a workaround over this.

auhlig:
I completely agree with @vinodborole.

Although we should think about the resolver. It should be easier to override.

bcornelis:
I definitely agree with  @vinodborole and @auhlig that properly configuring the Openstack installation should be the first thing to do, but unfortunately, it's configured by external companies, so it's not an option for now.
Besides that I share the same reasoning as @auhlig that anyway the library currently doesn't allow for any modification to this class, which can be improved.
@auhlig: can you agree to solve this by 1) using the API's to expose the service or 2) include it in the Config class? I can update the patch to match your decision.
Thanks!

auhlig:
I'd rather go for the config option. If you already have something in mind: Could you share an example of how it would be used @bcornelis ?

bcornelis:
It would be used like `.withConfig(Config.newConfig().withEndpointURLResolver(resolver))`.

auhlig:
Ah even better if that patter remains. Looking forward to the updated commit. Could you also add a short test for that? Thanks in advance for your work. Feel free to reach out in case you need help.

bcornelis:
done! Please let me know what you think about it.

auhlig:
Very cool @bcornelis. What's your opinion @vinodborole?

vinodborole:
LGTM, a documentation around this will help a lot, @bcornelis  can you also update the doc on the website at these places

http://openstack4j.com/learn/identity/identity-v2/#services-and-endpoints

http://openstack4j.com/learn/identity/identity-v3/#services-and-endpoints

what do you think @auhlig ?

auhlig:
Good idea @vinodborole. 
Since this is not specific to idv2/v3 I was thinking of a short section in http://openstack4j.com/learn/getting-started guide explaining  config options. Maybe @bcornelis can start with the new resolver and I volunteer to complete this. Makes sense?

We now have the old `.withResolver(resolver)` and the new `.withEndpointURLResolver(resolver))`. Mark the old as deprecated or replace with the new option? Or would there still be a use-case for this?

bcornelis:
@auhlig,
I suggest to leave the withResolver logic as it allows for easy changing the ServiceVersionResolver. Although, I suggest to rename it to something like `.withServiceVersionResolver(resolver)`.

I'll provide you with an initial version of the getting started documentation for this addition. Can you please give me a hint how/where to update documentation? Or I just send you my suggestion and you'll incorporate it?

auhlig:
Sounds good.
Regarding the docs: I'll prepare a section for the config options during the day and notify you when it's done.

bcornelis:
@auhlig Were you already able to prepare the docs? Thanks!

auhlig:
Thanks for the reminder and sorry for the delay.
No pushing in the evening. Will do it tomorrow morning.

auhlig:
@bcornelis Maybe you can mention the endpoint resolver here https://github.com/ContainX/openstack4j.com/commit/76448629369fad41aa82c55f10b087bb57a254e3#diff-2158f2689ed69342e8755af3e6844928R243 ?

auhlig:
**push** @vinodborole @bcornelis

vinodborole:
@auhlig i guess we should do it as part of another PR, better for tracking that way.

auhlig:
Makes sense. Then I guess this one can be merged.

Issue #864, opened 2016-10-24
the NeutronFloatingIP class miss the status property


whaon:
```
{
    "floatingip": {
        "floating_network_id": "376da547-b977-4cfe-9cba-275c80debf57",
        "router_id": "d23abc8d-2991-4a55-ba98-2aaea84cc72f",
        "fixed_ip_address": "10.0.0.3",
        "floating_ip_address": "172.24.4.228",
        "project_id": "4969c491a3c74ee4af974e6d800c62de",
        "tenant_id": "4969c491a3c74ee4af974e6d800c62de",
        "status": "ACTIVE",
        "port_id": "ce705c24-c1ef-408a-bda3-7bbd946164ab",
        "id": "2f245a7b-796b-4f26-9cf9-9e82d248fda7"
    }
}
```
NeutronFloatingIP class misss the **status** property

auhlig:
Would you like to submit another PR @whaon ?

auhlig:
PR merged. Can we close?

PR #866, opened 2016-10-25
add missing propety: status

whaon:
fix issue:#864


auhlig:
And another one.. Many thanks @whaon, @tongh-qtl

PR #871, opened 2016-10-28
fix json mapping for subnetId

atkinart:
http://developer.openstack.org/api-ref-compute-v2.1.html

`"ip_address": "192.168.1.3",
  "**subnet_id**": "f8a6e8f8-c2ec-497c-9f23-da9616de54ef"`

class org.openstack4j.openstack.compute.domain.NovaInterfaceAttachment.NovaFixedIp


auhlig:
Thanks for your contribution @atkinart. LGTM.
Since this doesn't look tested so far: Could you add a short one? Thanks in advance.

auhlig:
Merging to get this in before the release.

Issue #874, opened 2016-11-01
Does Openstack4J support  Block Storage API V3 Backups (backups)


easyit:
A backup is a full copy of a volume stored in an external service.  Please refet to http://developer.openstack.org/api-ref/block-storage/v3/?expanded=list-backups-with-detail-detail

Does Openstack4J support below functions ?
List backups with detail.
Show backup detail
Delete a backup
Restore a backup
Create a backup
List backups for tenant

auhlig:
I'm not aware of this in os4j. Are you @vinodborole?

You're very welcome to contribute and submit a PR @easyit. Feel free to reach out in case you need help.

vinodborole:
Its not there i guess, yes please go ahead and contribute. let us know if you need any help around. :+1:

easyit:
Ok, I will work on it

aiionx:
Its there already
https://github.com/ContainX/openstack4j/commit/90ea00add7cac534146a0c9fe6581ebeb73f184c

But snapshot_id is not supported it seems
The parameters you can send are:
```
    {"backup":
       {
          "force": false, 
          "description": null, 
          "snapshot_id": "b4b3258d-555a-4fce-8f53-69cc2ae96d3c", 
          "incremental": false, 
          "volume_id": "bf20730c-ddcb-4935-8b23-2ae7318bd274", 
          "container": null, 
          "name": null
       }
    }
```

aiionx:
Creating a volume backup from a Snapshot, which is supported in v2, is not implemented yet.
It is already implemented and supported in the openstack CLI.

Example:

```(openstack) volume backup create bf20730c-ddcb-4935-8b23-2ae7318bd274 --snapshot b4b3258d-555a-4fce-8f53-69cc2ae96d3c```

### 1. Lets get the volume data

REQ: 
```
$ curl -g -i -X GET \ 
  http://10.89.138.75:8776/v2/d1407d0e84e543a6a7ad9a31a5c65358/volumes/bf20730c-ddcb-4935-8b23-2ae7318bd274 \ 
  -H "User-Agent: python-cinderclient" -H "Accept: application/json" \
  -H "X-Auth-Token: {SHA1}token"
```

RESP: 
```
[...]

PR #898, opened 2016-12-02
support snapshot_id in volume backups

aiionx:
Create a volume backup from a snapshot.
This feature is not documented in 
http://developer.openstack.org/api-ref/block-storage/v3/?expanded=create-a-backup-detail
but it is indeed supported by the Openstack server and the CLI

https://github.com/ContainX/openstack4j/issues/874


auhlig:
Cool. Thanks @aiionx. Looks good from my side. 
Your turn @vinodborole :)

vinodborole:
LGTM

Issue #896, opened 2016-11-25
Unable to create alarm in aodh alrming service using openstack4j


zhangjianweibj:
![image](https://cloud.githubusercontent.com/assets/18585129/20618519/781d57e8-b32a-11e6-845d-a882a4d3ec6a.png)
 
openstack-n,create alarm,the request is sent to 8777 port, which is telemetry metering service. in fact it should be sent to 8042 port, where the telemetry alarming service is running.

zhangjianweibj:
![image](https://cloud.githubusercontent.com/assets/18585129/20618607/10763c6c-b32b-11e6-8a6e-da62b2fc46a8.png)

auhlig:
PR merged. Please close if the issue is fixed @zhangjianweibj.

PR #900, opened 2016-12-06
Openstack telemetry aodh support

zhangjianweibj:
telemetry_aodh support


zhangjianweibj:
fix https://github.com/ContainX/openstack4j/issues/896

zhangjianweibj:
the new feature documented in http://docs.openstack.org/developer/aodh/, which is designed to scale telemetry alarm service,but this new feature is not supported by openstack4j .

auhlig:
Hi @zhangjianweibj ,
Thanks for filing the issue and creating a PR.
Before we can take this further, I'd like to ask you to add some tests.
Feel free to reach out in case you need help.

zhangjianweibj:
hi @auhlig ,telemetry aodh test cases already commited

core-test/src/main/java/org/openstack4j/api/telemetry/CapabilitiesV3Test.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    protected Service service() {
+        return Service.TELEMETRY;
+    }
+
+    @Test
+    public void getSampleTest() throws IOException {
+        respondWith(JSON_CAPABILITIES);
+        Capabilities capabilities = osv3().telemetry().capabilities().get();
```
auhlig: Shouldn't this be like `.get(<id>)` ? Else maybe `.list()` is be more consistent, or?

auhlig: Maybe also check at at least one attribute.

core/src/main/java/org/openstack4j/openstack/telemetry/internal/BaseTelemetryAodhServices.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+
+public class BaseTelemetryAodhServices extends BaseOpenStackService {
+
+	protected BaseTelemetryAodhServices() {
+		super(ServiceType.TELEMETRY_AODH, EndpointFunction.instance);
+	}
+	
+
+	private static class EndpointFunction implements Function<String, String> {
```
auhlig: Could you briefly describe why this is necessary?
Usually OpenStack services get the endpoint from the service catalog.

vinodborole:
i agree @auhlig

core/src/main/java/org/openstack4j/openstack/telemetry/internal/BaseTelemetryAodhServices.java
zhangjianweibj: yeah,OpenStack services can get the endpoint from the service catalog.but openstack4j is a tool for  developers.users can get service information  from osclient.BaseTelemetryAodhServices is responsible for the related work.

auhlig: Understood. I don't really like doing that here. Maybe we could the include that in the path constant instead. Does that make sense?

zhangjianweibj: thank for your suggestion,path constant maybe a good idea at present.but,telemetry service endpoint may be changed with version information or not in the future. EndpointFunction solution can work with that situation.  another solution we can do like
image

what is your idea,auhlig?

zhangjianweibj: image

PR #914, opened 2017-01-08
Add api for nova service list

esommar:
Api implementation for nova service-list. Test included.

### DIFF
--- a/core/src/main/java/org/openstack4j/api/compute/ext/ServicesService.java
+++ b/core/src/main/java/org/openstack4j/api/compute/ext/ServicesService.java
@@ -1,0 +1,23 @@
+package org.openstack4j.api.compute.ext;
+
+import java.util.List;
+
+import org.openstack4j.model.compute.ext.Service;
+
+/**
+ * API which supports the "os-services" extension.
+ *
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/openstack4j/api/compute/ext/ServicesService.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.List;
+
+import org.openstack4j.model.compute.ext.Service;
+
+/**
+ * API which supports the "os-services" extension.
+ *
+ * @author Stephan Latour
+ */
+public interface ServicesService{
```
gondor: Could you prefix a space before the brace

gondor:
Syntax comment, otherwise LGTM

core/src/main/java/org/openstack4j/api/compute/ext/ServicesService.java
esommar: Thanks, done!

auhlig:
Cool. LGTM. Many thanks @esommar and @gondor :)

Issue #903, opened 2016-12-13
Unable to remove the last route from the Router


silvanstrueby:
I have a static route on my router.
As far as I can see, unfortunately it's impossible to delete the last route from a router with this API.
```
Router router = client.networking().router().get("ROUTER_ID");
router.getRoutes().clear();
client.networking().router().update(router);
```
Because if there are no routes left in the router, the body of the router-update-request is getting parsed without the attribute routes.
```
{
  "router" : {
    "name" : "NAME",
    "external_gateway_info" : {
      "network_id" : "ID",
      "enable_snat" : true
    },
    "admin_state_up" : true,
  }
}
```
That's why the routes won't get updated.
It would work if the body is getting parsed like this.
```
{
  "router" : {
    "name" : "NAME",
    "external_gateway_info" : {
      "network_id" : "ID",
      "enable_snat" : true
    },
    "admin_state_up" : true,
    "routes" : []
  }
}
```
Even when the JavaDoc says update does only work on [ name, admin_state_up and external_gateway_info ], I don't see any other option to remove static routes.

Do I miss something here or is there another possibility to remove static routes?

auhlig:
Hey @silvanstrueby ,
Is this missing in OpenStack iitself? Or do we need to update OpenStack4j? If so, you're very welcome to submit a PR.

silvanstrueby:
Hi @auhlig,
I would like to create a PR with my fix.
I tried to push my branch but obviously I'm not entitled to do so. Could you please give me a quick hint how i could create a PR?
Thx

vinodborole:
@silvanstrueby 
You push to your forked copy and create a Pull request from the git gui;

auhlig:
PR merged. Thanks a lot for your contribution @silvanstrueby .
Can the issue  be closed?

PR #917, opened 2017-01-10
Issue 903 remove last route

silvanstrueby:
Bugfix for issue #903.


auhlig:
Many Thanks @silvanstrueby.
Could you add a short test to verify your fix works?

silvanstrueby:
Hi @auhlig 

I tried to implement a test but I'm afraid to say that unfortunately it would take me too long to write reasonable tests. Because I'm not familiar with your test setup and I couldn't find any existing tests for the RouterService to get some inspiration.

The fix itself compared to the effort I would need to test it is disproportionally in my opinion.
I know it doesn't really help you, but we use this fix in our productive system and it solved the problem.

I hope you can understand my point of view?

auhlig:
Mh. What do you think on this @vinodborole?

vinodborole:
@auhlig should not be a problem LGTM

PR #945, opened 2017-02-24
added unit test readme

auhlig:
Since users asked where the unit tests are and how they work.
Also updated `Currently we support a minimum version of Java 6.` to `.. Java 7.`
Are you okay with the changes @vinodborole, @gondor?


vinodborole:
awesome! this was indeed required. Thanks a lot @auhlig, LGTM

PR #947, opened 2017-02-27
fix domain null pointer exception

ayberkcal:


auhlig:
Cool. Thanks @ayberkcal.

Issue #795, opened 2016-09-02
Failed to get console log 


Francesko90:

Francesko90:
Follow up

Francesko90:
Possible connection https://github.com/ContainX/openstack4j/issues/839

lorenzo-biava:
This still happens with the `openstack4j-resteasy` connector, v3.0.3.

It looks like Resteasy client is looking for a `MessageBodyWriter` for the `String` type (the console log API is one of the few that is implemented with a JSON String instead of a POJO).
However, the only two [registered providers](https://github.com/ContainX/openstack4j/blob/3.0.3/connectors/resteasy/src/main/java/org/openstack4j/connectors/resteasy/ResteasyClientFactory.java#L29) are `JacksonJsonProvider` and `InputStreamProvider`.
The latter is obviously skipped and the `JacksonJsonProvider` is also unable to write a `String` body (see the check being made by the provider - we are currently using [v2.3.2](https://github.com/FasterXML/jackson-jaxrs-providers/blob/jackson-jaxrs-providers-2.3.2/base/src/main/java/com/fasterxml/jackson/jaxrs/base/ProviderBase.java#L539), but should be the same for the version [v2.7.3](https://github.com/FasterXML/jackson-jaxrs-providers/blob/jackson-jaxrs-providers-2.7.3/base/src/main/java/com/fasterxml/jackson/jaxrs/base/ProviderBase.java#L535) defined in the connector's POM).

I'm currently not able to test it using a newer version of the `JacksonJsonProvider`, but looking at the code  it seems like the behavior is the same.

@auhlig Could you let me know if you think it is just a problem with the Jackson library version or it is a problem to be fixed in the client configuration (say adding a `MessageBodyWriter` for the JSON String type in the `ResteasyClientFactory`)? Thanks.

auhlig:
Hi @lorenzo-biava,
Many thanks for investigating.
I hoped it's just a problem with the jackson lib. that would be resolved for us by upgrading.
As you describe it that might not be the case. 
How much work would it be to fix it in os4j?

lorenzo-biava:
Well I think it would be quite fast, using one of two options:
1) As far as i can see, the JSON String way (i.e. the `BaseOpenStackService.Invocation#json` method) is only used by few APIs. It might be removed and replaced with standard entity encoding (which I personally think would be much more consistent and readable).
Indeed, I temporarily solved the console log problem defining a proper POJO:
```java
[...]

PR #949, opened 2017-02-28
Fix get console output not working with Resteasy connector (#795)

lorenzo-biava:
This PR replaces the JSON String encoding of the compute.getConsoleOutput API request with a standard `EntityModel `so that it fixes issue #795.

With this entity modification, it also changes the default console length, which is now the full console instead of 50 lines.

Some unit tests has been added for this API.


lorenzo-biava:
@auhlig It's ready for me. Let me know if there's anything that needs improvement...

auhlig:
Thanks for all the work @lorenzo-biava.
Could you explain the need for the `takeRequest()` method in AbstractTest?
It works without and if necessary, you can always use `server.takeRequest()` in the respective test classes that extend AbstractTest.
Besides this looks good.

lorenzo-biava:
Sure. I wanted to make sure that the console output lenght logic was correct and so I needed to check the request for the `length` parameter presence/absence.
So I added a common `takeRequest()` method for a couple of reasons, even though it is not absolutely necessary.

The first is that I thought that an abstraction on the mock server like `respondWith()` would be more consistent (and perhaps the `takeRequest()` method might have additional logic in the future).

The second, more practical one, is that in the `createServer` test there's already a variable named `server`, so it would clash with the mock server field.

I can change the variable in the `createServer()` method or use `this.server.takeRequest()` if you prefer, but i felt it would be less confusing having a dedicated common method to access common mock server functionalities indirectly.

auhlig:
Makes sense. Thanks @lorenzo-biava. Merging now.

Issue #920, opened 2017-01-16
expose os-vol-host-attr:host in Volume model


easyit:
As mentiond in [http://developer.openstack.org/api-ref/block-storage/v3/?expanded=list-accessbile-volumes-with-details-detail](url) ,  the os-vol-host-attr:host	contains current back-end of the volume. Could we add extra interface in the Volume model as below and return this info ?
`String getBackend()`

auhlig:
Hi @easyit, Sure. You know the drill :) Waiting for your PR.

Issue #901, opened 2016-12-09
Some attributes for CinderVolume and Hypervisor are not supported


piotrroztkowski:
CinderVolume attribute os-vol-host-attr:host is available through REST API but not supported by OpenStack4J. Hypervisor (os-hypervisor) attributes state and status are available through REST API but not supported by OpenStack4J

auhlig:
Hey @piotrroztkowski ,
Thanks for reporting.
Would you like to contribute and submit a PR to address this?

piotrroztkowski:
Hi,

Yes, we can contribute and address these issues but we need some time to create development environment for OpenStack4j.

auhlig:
That's great! 
Looking forward to your PR.
Let us know in case you need help.

PR #952, opened 2017-03-01
add missed 'os-vol-host-attr:host' property on CinderVolume

ayberkcal:
Releated these issues
https://github.com/ContainX/openstack4j/issues/920
https://github.com/ContainX/openstack4j/issues/901


auhlig:
Thanks @ayberkcal.

Issue #944, opened 2017-02-22
osclient.objectStorage().objects().get(containerName, name) doesn't seem to parse ETag


gonewest818:
Using the version 3.0.3, I am getting different swift objects from `osclient.objectStorage().objects().list(containerName)` versus `osclient.objectStorage().objects().get(containerName, name)` in that the former gives me a list of SwiftObjectImpl containing ETag, wheres the latter gives me a single SwiftObjectImpl with no ETag. In this particular use case I only want the ETag for the individual file I am downloading with `get`, so it's unfortunate that the ETag isn't there.

The SwiftObjectImpl in the `get()` case is being built in an apply method on the ParseObjectFunction:

    @Override
    public SwiftObject apply(HttpResponse resp) {
        return SwiftObjectImpl.builder()
                  .name(location.getObjectName())
                  .containerName(location.getContainerName())
                  .mimeType(resp.header(CONTENT_TYPE))
                  .sizeBytes(asLong(resp.header(CONTENT_LENGTH)))
                  .metadata(MapWithoutMetaPrefixFunction.INSTANCE.apply(resp.headers()))
                  .lastModified(Parser.toRFC822DateParse(resp.header(LAST_MODIFIED)))
                  .build();

which includes explicit handling of content type and content length, but it seems to me it's just not handling etags.

auhlig:
Hi @gonewest818,
Thanks for filing the issue.
You're right. I guess we easily could add the ETag attribute to the `get()` method.
Do you want to contribute and submit a PR for this?
Would be much appreciated.
Feel free to reach out to us in case you need help

gonewest818:
Sure, I'll take a look at that.

gonewest818:
btw I've got the code fixed. I'm now trying to figure out how and where to add a unit test for this.

gonewest818:
closed.  thank you!

PR #958, opened 2017-03-04
Issue 944: etag support in os.objectStorage().objects().get()

gonewest818:
This PR resolves issue #944  by modifying ParseObjectFunction such that we now get the eTag for objects retrieved from swift.  Also added a unit test.


auhlig:
Great. Thanks for fixing this @gonewest818.

Issue #957, opened 2017-03-03
DownloadImage test failure in ImageV2Tests.java


gonewest818:
If I clone from master and "make test", the code builds and runs tests, but the tests stop with an assertion failure in the DownloadImage test. It's expecting a return code 400 but in my tests the result is 200.  Looks like the test case has been there for at least 6 months...  is this unique to me, or is the test actually broken?

auhlig:
Travis CI shows everything is green. Must be related to your environment.

gonewest818:
"make test" is running the pristine environment defined in your Dockerfile, which I can totally imagine must define a different environment than what's in Travis. Especially since the Dockerfile is using the maven:latest image and the contents would be changing over time.

In Travis, I see that particular test is throwing an exception (file not found) and returning 400.

I could rename this issue to be more clearly related to running tests in Docker... ?

auhlig:
Indeed https://travis-ci.org/ContainX/openstack4j/jobs/206935719#L2838.
Looks like the [unit test](https://github.com/ContainX/openstack4j/blob/master/core-test/src/main/java/org/openstack4j/api/image/v2/ImageV2Tests.java#L236) is covering both the 200 and 400 response.

gonewest818:
I think the unit test is failing in Docker because the default user is root. So when that test case tries to access URL("file:////test.iso") it actually can. It's not failing as expected.

gonewest818:
closed.  thank you!

PR #959, opened 2017-03-05
issue 957: run as non-root user in docker

gonewest818:
This PR resolves Issue #957 in which one of the unit tests fails because it has root privileges in docker.  With this patch maven runs in docker as a non-root user named "builder" instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,5 +2,10 @@ FROM maven:latest
 
 COPY . /usr/src/app
 WORKDIR /usr/src/app
+
+RUN useradd -m -u 1000 builder
+RUN chown -R builder /usr/src/app
+USER builder
+
 RUN mvn clean compile
 
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 IMAGE_NAME := openstack4j-dev$(if $(GIT_BRANCH),:$(GIT_BRANCH))
-DOCKER_RUN := docker run --rm -it --privileged $(IMAGE_NAME)
+DOCKER_RUN := docker run --rm -it --privileged -e MAVEN_CONFIG=/home/builder/.m2 $(IMAGE_NAME)
 
 default: build
 


auhlig:
Hey @gonewest818,
Thanks for working on this.
But I can still see it failing https://travis-ci.org/ContainX/openstack4j/jobs/207814369#L2845

gonewest818:
Well, in my understanding of the test case, I think that's a "success":

     public void DownloadImage() throws IOException {
         respondWith(200);
         String imageId = "4b434528-032b-4467-946c-b5880ce15c06";
         URI uri = null;
         try {
             uri = new URI("file:////test.iso");
         }catch (URISyntaxException e) {
             e.printStackTrace();
         }
         File file = new File(uri);
         ActionResponse download = osv3().imagesV2().download(imageId, file);
         // Should fail to write to file
         assertEquals(download.getCode(), 400);
     }

The `respondWith(200)`  mocks the [GET /images/{imageId}/file](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/image/v2/internal/ImageServiceImpl.java#L158) that is going to occur inside the `imagesV2().download()`.  In the Travis log you see the result of that mock

    INFO: MockWebServer[9292] received request: GET /v2/images/4b434528-032b-4467-946c-b5880ce15c06/file HTTP/1.1 and responded: HTTP/1.1 200 OK

Next, URI "file:////test.iso" is used to describe the file path where the downloaded bytes will be written. However /test.iso does not exist and cannot be written. This means [constructing the output stream](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/image/v2/internal/ImageServiceImpl.java#L163) throws an exception, which [is properly caught and the exception handler prints a stack trace](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/image/v2/internal/ImageServiceImpl.java#L175) like you see in Travis.

The exception handler in `imagesV2().download()` always returns 400. So the unit test wraps up with an assertion confirming the 400 and the test therefore "succeeds."

auhlig:
Makes sense. Thanks for fixing that @gonewest818.

PR #960, opened 2017-03-05
Gnocchi (Telemetry) support [Not ready, but review needed]

adyachok:
Enable [OpenStack Gnocch](https://wiki.openstack.org/wiki/Gnocchi)i (Telemetry) support.

### DIFF
--- a/core/src/main/java/org/openstack4j/api/OSClient.java
+++ b/core/src/main/java/org/openstack4j/api/OSClient.java
@@ -15,6 +15,7 @@ import org.openstack4j.api.senlin.SenlinService;
 import org.openstack4j.api.storage.BlockStorageService;
 import org.openstack4j.api.storage.ObjectStorageService;
 import org.openstack4j.api.tacker.TackerService;
+import org.openstack4j.api.telemetry.TelemetryGnocchiService;
 import org.openstack4j.api.telemetry.TelemetryService;
 import org.openstack4j.api.trove.TroveService;
 import org.openstack4j.api.types.Facing;
@@ -132,6 +133,13 @@ public interface OSClient< T extends OSClient<T>> {
     boolean supportsTelemetry();
 
     /**
+     * Determines if the Telemetry (Gnocchi) service is supported
+     *
+     * @return true if supports Telemetry
+     */
+    boolean supportsTelemetry_gnocchi();
+
+    /**
      * Determines if the Shared File Systems (Manila) service is supported
      *
      * @return true if supports Shared File Systems
@@ -206,6 +214,14 @@ public interface OSClient< T extends OSClient<T>> {
      * @return the telemetry service
      */
     TelemetryService telemetry();
+
+    /**
+     * Returns the Gnocchi Telemetry Service API
+     *
+     * @return the Gnocchi telemetry service
+     */
+    TelemetryGnocchiService gnocchiTelemetry();
+
 
     /**
      * Returns the Shared File Systems API
--- a/core/src/main/java/org/openstack4j/openstack/telemetry/domain/GnocchiMetricsMetadata.java
+++ b/core/src/main/java/org/openstack4j/openstack/telemetry/domain/GnocchiMetricsMetadata.java
@@ -1,0 +1,92 @@
+package org.openstack4j.openstack.telemetry.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.openstack4j.model.telemetry.Meter;
+
+import java.util.List;
+
+/**
+ * Created by eandgya on 3/2/17.
+ */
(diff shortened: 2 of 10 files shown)


auhlig:
Hi @adyachok,
Many thanks for contributing. We'll take a look.
Feel free to reach out to us in case you need any help.

core/src/main/java/org/openstack4j/openstack/telemetry/domain/GnocchiMetricsMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+            public String granularity;
+
+            @JsonProperty
+            public int points;
+
+            @JsonProperty
+            public String timespan;
+
+            @Override
+            public String toString() {
```
auhlig: I would prefer something like this https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/identity/v3/domain/KeystoneDomain.java#L90-L98

core/src/main/java/org/openstack4j/openstack/telemetry/domain/GnocchiMetricsMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+                return "Definition{" +
+                        "granularity='" + granularity + '\'' +
+                        ", points=" + points +
+                        ", timespan='" + timespan + '\'' +
+                        '}';
+            }
+        }
+
+        @Override
+        public String toString() {
```
auhlig: see previous comment

core/src/main/java/org/openstack4j/openstack/telemetry/domain/GnocchiMetricsMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+                    "aggregationMethods=" + aggregationMethods +
+                    ", backWindow=" + backWindow +
+                    ", definition=" + definition +
+                    ", name='" + name + '\'' +
+                    '}';
+        }
+    }
+
+    @Override
+    public String toString() {
```
auhlig: see previous comment

core/src/main/java/org/openstack4j/openstack/telemetry/internal/MetricGnocchiServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import org.openstack4j.openstack.telemetry.domain.GnocchiMetricsMetadata;
+
+import java.util.List;
+
+/**
+ * Created by eandgya on 3/3/17.
+ */
+public class MetricGnocchiServiceImpl extends BaseTelemetryGnocchiServices implements MetricGnocchiService {
+    @Override
+    public List<? extends Meter> list() {
```
auhlig: I guess that's WIP

core/src/main/java/org/openstack4j/openstack/telemetry/internal/MetricGnocchiServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        return null;
+    }
+
+    @Override
+    public void putSamples(List<MeterSample> sampleList, String meterName) {
+
+    }
+    @Override
+    public List<GnocchiMetricsMetadata> metricsMetadata() {
+        GnocchiMetricsMetadata[] meters = get(GnocchiMetricsMetadata[].class, uri("/v1/metric")).execute();
```
auhlig: Instead of introducing a new path (`/metric`) here I'd suggest to create a const in https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/core/transport/ClientConstants.java

core/src/main/java/org/openstack4j/api/OSClient.java
```diff
@@ -208,6 +216,14 @@
     TelemetryService telemetry();
 
     /**
+     * Returns the Gnocchi Telemetry Service API
+     *
+     * @return the Gnocchi telemetry service
+     */
+    TelemetryGnocchiService gnocchiTelemetry();
```
auhlig: Is this the Gnocchi Telemetry Service a separate OpenStack service that should live here or is it be part of the telemetry service (e.g. just different TSDB/backend to ceilometer)?

auhlig:
Still working on this @adyachok? Don't hesitate to reach out to us in case you need help.

pdube:
Hey I was reading some of the Gnocchi documentation and it seems to me that it doesn't really make sense to implement the `TelemetryService` as most of it is empty. Creating your own `MetricService` would enable it to have it's own set of functionality. As it stands, there are a lot of potential NPEs if you try to use the TelemetryGnocchiService like the CeilometerTelemetryService

Issue #948, opened 2017-02-28
Parsing error while listing clusters : Failed to parse Date value


sunayanasarangi:
Hallo Team,
I have an apache spark cluster on openstack sahara and I want to use the openstack4j api. I have established a V2 connection. I am able to list flavours, plugins, servers etc. 
But with functions like os.sahara().clusters().list() , os.sahara().clusters().get(clusterId) , os.sahara().nodeGroupTemplates().list(), os.sahara().nodeGroupTemplates().list(), os.sahara().clusterTemplates().list() or os.sahara().clusterTemplates().get(templateId) I am getting a parsing error. Am I missing something ? Any help will be greatly appreciated !!

Below is my code 
public static void main(String[] args) {
		//Creating and authenticating against OpenStack
		OSClientV2 os = OSFactory.builderV2()
				.endpoint("http://asflexsm.informatik.uni-stuttgart.de:5000/v2.0")
				.credentials("sunayana.sarangi", "*******")
				.tenantName("FlexMash")
				.authenticate();
           List<? extends Cluster> clusters = os.sahara().clusters().list();
           System.out.println(clusters);
}

pom.xml has below dependency:
<dependency>
  		<groupId>org.pacesys</groupId>
  		<artifactId>openstack4j</artifactId>
  		<version>3.0.0</version>
  		<classifier>withdeps</classifier>
	</dependency>

Error log: 
Exception in thread "main" javax.ws.rs.ProcessingException: Error reading entity from input stream.
	at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:889)	at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:808)	at org.glassfish.jersey.client.ClientResponse.readEntity(ClientResponse.java:326)
	at org.glassfish.jersey.client.InboundJaxrsResponse$1.call(InboundJaxrsResponse.java:115)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:228)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:419)
	at org.glassfish.jersey.client.InboundJaxrsResponse.runInScopeIfPossible(InboundJaxrsResponse.java:267)
[...]

Issue #432, opened 2015-07-23
Sahara response date-time stamps cause exception


bgombert:
It appears that the opensatck4j support for Data Processing(sahara) has an issue with rest api responses that contain a date-time stamp (example below). Debugging was turned on and the cluster creation is successful however the databinding of the response fails. It seems that the openstack sahara service uses a different format for datetime stamps. Seems that the datetime stamps may not be ISO 8601 conformant. Debugging api calls to nova shows datetime stamps that are ISO 8601 conformant.

Have others had success using the openstack4j sahara api? If so what version of openstack are you using?

openstack envirnment:
            devstack kilo with sahara enabled

pom contains:
              dependency for openstack4j version 2.0.4, withdeps

java code for cluster creation:
    Cluster cluster = os.sahara().clusters().create(Builders.cluster()
                                 .name("vanilla-cluster").pluginName("vanilla").hadoopVersion("1.2.1")
                                 .template("cb6b597c-6a6c-4186-8c1b-a6db0419ed56")
                                 .image("8cf7eb97-ae07-45a6-a73f-322dd839972b")
                                 .keypairName("mykey").build());

Jul 21, 2015 4:18:48 PM org.glassfish.jersey.filter.LoggingFilter log
INFO: 1 \* Sending client request on thread main
1 > POST http://1.1.1.1:8386/v1.1/1c9e77bd7f5146eaaf64bf1ed053b73d/clusters
1 > Accept: application/json
1 > Content-Type: application/json
1 > User-Agent: OpenStack4j / OpenStack Client
1 > X-Auth-Token: 4d640e13f6844cabb5ea5b4bed3b9556
{
  "name" : "vanilla-cluster",
  "cluster_template_id" : "cb6b597c-6a6e-4186-8c1a-a6db0419ed56",
  "default_image_id" : "8cf7eb07-ae07-4596-a73f-322dd839972b",
  "user_keypair_id" : "barry",
  "plugin_name" : "vanilla",
  "hadoop_version" : "1.2.1"
}

Jul 21, 2015 4:18:50 PM org.glassfish.jersey.filter.LoggingFilter log
INFO: 2 \* Client response received on thread main
2 < 202
2 < Connection: keep-alive
2 < Content-Length: 2510
2 < Content-Type: application/json
2 < Date: Tue, 21 Jul 2015 20:14:52 GMT
2 < X-Openstack-Request-Id: req-c872e1ad-f355-4c0e-892a-1d1c70f00311
[...]

Issue #423, opened 2015-07-14
Error reading entity from input stream


niebloomj:
I am getting an error when I try to list sahara clusters or create a sahara cluster. It is an internal error with parsing a date that contains a T apparently. I am running openstack4j 2.0.3 and devstack Kilo

My Code:

Cluster cluster = manager.client
        .sahara()
        .clusters()
        .create(Builders
                .cluster()
                .name("test-cluster19")
                .pluginName("vanilla")
                .hadoopVersion("2.6.0")
                .image(imageId)
                .addNodeGroup(
                        Builders.nodeGroup().name("worker").flavor("1")
                                .addNodeProcess("namenode")
                                .addNodeProcess("resourcemanager")
                                .addNodeProcess("historyserver")
                                .addNodeProcess("oozie")
                                .addNodeProcess("nodemanager").count(1)
                                .build())
                .addNodeGroup(
                        Builders.nodeGroup().name("worker2")
                                .flavor("1").addNodeProcess("datanode")
                                .count(3).build())
                // .keypairName("mykeypair")
                // .managementNetworkId("")
                .build());

The exception:

Exception in thread "main" javax.ws.rs.ProcessingException: Error reading entity from input stream.
    at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:866)
    at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:783)
    at org.glassfish.jersey.client.ClientResponse.readEntity(ClientResponse.java:326)
    at org.glassfish.jersey.client.InboundJaxrsResponse$1.call(InboundJaxrsResponse.java:111)
    at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
    at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
    at org.glassfish.jersey.internal.Errors.process(Errors.java:228)
    at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:399)
    at org.glassfish.jersey.client.InboundJaxrsResponse.readEntity(InboundJaxrsResponse.java:108)
    at org.openstack4j.connectors.jersey2.HttpResponseImpl.readEntity(HttpResponseImpl.java:112)
    at org.openstack4j.core.transport.HttpEntityHandler.handle(HttpEntityHandler.java:56)
    at org.openstack4j.core.transport.HttpEntityHandler.handle(HttpEntityHandler.java:19)
[...]

PR #961, opened 2017-03-06
Change date-time format in Sahara API to ISO 8601

ekasitk:
The date-time format in Sahara API has been changed to ISO 8601, as reported by #423, #432, and #948. This pull request is to update Sahara classes in openstack4j accordingly.


auhlig:
Hi @ekasitk,
Thanks for tackling this.
Could you at least add one test for this? Although I think this should work

PR #962, opened 2017-03-06
add public host property getter

ayberkcal:


auhlig:
Thanks for tackling that @ayberkcal.

Issue #963, opened 2017-03-06
PortListOptions with port name is not working


evgenyfedoruk:
Listing neutron ports with PortListOptions where displayName() was set - not working properly. All ports are returned instead of ports with the specified name.

Code example:
List<? extends Port> ports = this.osClient.networking().port().list(PortListOptions.create().displayName(portName));

The issue is in class org.openstack4j.model.network.options.PortListOptions
method - public PortListOptions displayName(String displayName)

"name" should be added to the queryParams map instead of "display_name"

auhlig:
Since this was merged, can you close this issue @evgenyfedoruk?

evgenyfedoruk:
The issue was fixed by deprecating the old method and adding a new one which works

PR #965, opened 2017-03-08
Fixing port name list option.

evgenyfedoruk:
"name" should be specified instead of "display_name".
Link to the issue: https://github.com/ContainX/openstack4j/issues/963

### DIFF
--- a/core/src/main/java/org/openstack4j/model/network/options/PortListOptions.java
+++ b/core/src/main/java/org/openstack4j/model/network/options/PortListOptions.java
@@ -61,13 +61,26 @@ public class PortListOptions {
     }
     
     /**
-     * The port name
-     * 
-     * @param displayName the port name
+     * DEPRECATED: This option is not valid anymore since
+     * neutron port does not have attribute named "display_name".
+     * Use {@link #name(String)} to filter ports by name
+     *
+     * @param displayName the port display name
      * @return options
      */
+    @Deprecated
     public PortListOptions displayName(String displayName) {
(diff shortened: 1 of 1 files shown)


auhlig:
Hi @evgenyfedoruk,
Thanks investigating and submitting this PR.
As the attribute was `display_name` before and changed to `name` later, would it make sense to have both methods in our lib? So how about keeping the displayName method and just adding the following?
```
public PortListOptions name(String name) {
        return add("name", name);
    }
```

evgenyfedoruk:
Hi @auhlig
You right, pushed an update with new method and old method deprecated.

core/src/main/java/org/openstack4j/model/network/options/PortListOptions.java
```diff
@@ -61,13 +61,26 @@ public PortListOptions adminState(boolean adminState) {
...
+     * neutron port does not have attribute named "display_name".
+     * Use {@link #name(String)} to filter ports by name
+     *
+     * @param displayName the port display name
      * @return options
      */
+    @Deprecated
     public PortListOptions displayName(String displayName) {
-        return add("display_name", displayName);
+        return add("name", displayName);
```
auhlig: And this one stays `"display_name"`, not?

evgenyfedoruk: Oh.. sure.

auhlig:
Many thanks for contributing @evgenyfedoruk.

Issue #966, opened 2017-03-12
Neutron Port implementation does not support port binding extension


evgenyfedoruk:
Neutron Port implementation does not support port binding data extension.
This extension should be supported for allowing users to create Neutron Ports with binding data.

This issue created for adding this extension support.

auhlig:
Since this was merged, can you close this issue @evgenyfedoruk?

evgenyfedoruk:
Port binding data was added to a port

PR #967, opened 2017-03-12
Adding neutron port binding data support.

evgenyfedoruk:
Adding binding extension data to neutron port
Link to an issue: https://github.com/ContainX/openstack4j/issues/966


auhlig:
Many thanks for contributing @evgenyfedoruk.

PR #972, opened 2017-03-16
Allow set ID when creating image

baseman79:
Currently if you set Id with ImageBuilder it will get ignored. This
prevents user passing custom image UUID for newly create image.


core/src/main/java/org/openstack4j/openstack/image/domain/functions/ImageForUpdateToHeaders.java
```diff
@@ -49,6 +50,7 @@ public static ImageForUpdateToHeaders instance() {
 
         Map<String, Object> headers = Maps.newHashMap();
 
+        addIfNotNull(headers, ID, from.getName());
```
auhlig: Shouldn't it be `from.getId()`?

baseman79: doh, thanks

auhlig:
Merged. Thanks again @baseman79.

PR #976, opened 2017-03-20
support ipv6 properties in subnet again

taemin77:
Hi @auhlig 
I couldn't find subnet ipv6 properties in the lastest snapshot.
I don't know what is wrong.
Could you check it again?
Once, I created PR again. 

ipv6_address_mode
: The IPv6 address modes specifies mechanisms for assigning IP
addresses. Value is slaac, dhcpv6-stateful, dhcpv6-stateless or null.
ipv6_ra_mode
: The IPv6 router advertisement specifies whether the networking service
should transmit ICMPv6 packets, for a subnet. Value is slaac,
dhcpv6-stateful, dhcpv6-stateless or null.


auhlig:
LGTM

PR #953, opened 2017-03-02
support updating enabledhcp in subnet

liujunpengwork:
support updating enabledhcp in subnet

Issue #973, opened 2017-03-17
Security Group Update


ayberkcal:
Hi,
I would like to update security group name and description but I can not find implementation in openstack4j. Just implemented get/delete/create/list. Am I missing something or update do not implemented?
Thank you.

auhlig:
Do you mean [this](https://github.com/ContainX/openstack4j/blob/6c3451d7073282535d5982b6bcc72eba28b7d0ba/core/src/main/java/org/openstack4j/api/networking/SecurityGroupService.java)?

ayberkcal:
Yes, that is what I am saying :) Thank you I will implement as soon as possible.

auhlig:
Cool. Looking forward :)

PR #978, opened 2017-03-20
Implement Security Group Update API & Support Enable DHCP

ayberkcal:
Implement missed implementation about security group update API and write test case https://github.com/ContainX/openstack4j/issues/973. And also fix https://github.com/ContainX/openstack4j/pull/953 syntax problem.


auhlig:
Thanks @ayberkcal.

Issue #974, opened 2017-03-17
List Security Group Rules According to Security Group Id


ayberkcal:
Hi,
I would like to list security group rules according to security group id. List action implemented but it does not have "field=value" parameter [*] [**]. Can I miss something?

[*] OpenStack API: https://developer.openstack.org/api-ref/networking/v2/?expanded=delete-security-group-detail,list-security-group-rules-detail#list-security-group-rules
[**] openstack4j SDK JavaDoc: http://openstack4j.com/javadoc/org/openstack4j/openstack/networking/internal/SecurityGroupServiceImpl.html

auhlig:
True. Would you like to implement this?

ayberkcal:
I am working on it.

PR #981, opened 2017-03-21
add security group and rules filterable list method

ayberkcal:
releated https://github.com/ContainX/openstack4j/issues/974


auhlig:
LGTM. Thanks @ayberkcal.

PR #984, opened 2017-03-24
Fix for correcting return result type in Magnum Container APIs

sohaninfo:
This checkin contains a smal fix. It corrects return result types in Magnum Container APIs.


auhlig:
Merged. Thanks @sohaninfo.

PR #986, opened 2017-03-31
Allow passing additional properties for Glance V2

baseman79:
Currently, there is no way to passing additional properties (e.g. hw_vif_model, hw_disk_bus) when creating image through Glance V2
https://developer.openstack.org/api-ref/image/v2/index.html?expanded=create-an-image-detail#create-an-image


auhlig:
LGTM.

auhlig:
Thanks for contributing @baseman79!

Issue #987, opened 2017-03-31
Allow Port Creation in Bulk


Rizwan-Qamar:
Allow Creating Ports in One Request
https://developer.openstack.org/api-ref/networking/v2/?expanded=bulk-create-ports-detail#bulk-create-ports

PR #988, opened 2017-03-31
Allow Port Creation in Bulk

Rizwan-Qamar:
#987


auhlig:
LGTM. Thanks for contributing @Rizwan-Qamar.

PR #989, opened 2017-04-03
Add additional properties of block device mapping api v2

baseman79:
- disk_bus
- device_type

Refers to [Link](https://docs.openstack.org/developer/nova/block_device_mapping.html#block-device-mapping-v2)


auhlig:
Apologies for the delay.
Thanks for contributing @baseman79

PR #998, opened 2017-04-19
Add client certificate for keystone tokenless

ningyougang:
Currently, keystone supports client certificate without having to issue a
token, so it is necessary to add client certificate to support keystone
tokenless feature.

### DIFF
--- a/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneAuthenticationTests.java
+++ b/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneAuthenticationTests.java
@@ -4,15 +4,21 @@ import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
+import okhttp3.mockwebserver.RecordedRequest;
 import org.openstack4j.api.AbstractTest;
+import org.openstack4j.api.OSClient;
 import org.openstack4j.api.OSClient.OSClientV3;
 import org.openstack4j.api.exceptions.RegionEndpointNotFoundException;
+import org.openstack4j.core.transport.Config;
 import org.openstack4j.model.common.Identifier;
 import org.openstack4j.model.identity.AuthVersion;
 import org.openstack4j.model.identity.v3.User;
 import org.openstack4j.openstack.OSFactory;
+import org.openstack4j.openstack.identity.v3.domain.KeystoneToken;
 import org.testng.annotations.Test;
 
 import com.google.common.collect.ImmutableMap;
@@ -349,6 +355,26 @@ public class KeystoneAuthenticationTests extends AbstractTest {
 
     }
 
+    /**
+     * check headers whether right from request
+     *
+     * @throws Exception
+     */
+    public void pass_headers_Test() throws Exception {
 
+        respondWith(JSON_USERS);
+
+        Map header = new HashMap();
+        String key = "X-Domain-Id";
+        String value = "default";
+        header.put(key, value);
+        KeystoneToken token = new KeystoneToken();
+        token.setEndpoint(authURL("/v3"));
+        OSClient.OSClientV3 osClient = OSFactory.clientFromToken(token).headers(header);
+        osClient.identity().users().list();
+
+        RecordedRequest request = server.takeRequest();
+        assertEquals(request.getHeader(key), value);
+    }
 
 }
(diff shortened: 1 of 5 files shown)


auhlig:
Cool. Could you add a test @ningyougang?

ningyougang:
Currently, i don't add the test. wait for me.

ningyougang:
@auhlig ,i am investigating the test, for example:https://github.com/ContainX/openstack4j/blob/master/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneAuthenticationTests.java#L62
it compared the userId or projectId from Token object with the prepared projectId.
But here, because of X509 protocol, there has no token info.
So we need to add test case for X509 here?
if need, can you give some suggestion for me?

junoyoon:
@ningyougang can we customize okhttp3 mockserver sth like following?

http://stackoverflow.com/a/41659253

liuhongjiang:
From my point of view, you need to know how the mockwebserver support client certification, and when making request in the test case with client certification, you need use mockwebserver takeRequest to get the request, and then verify the client certification in the request, if the certification is verified, then return the response.

* [mockwebserver takeRequest](https://github.com/square/okhttp/blob/master/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java#L240)

liuhongjiang:
I find how to get client certification here: http://stackoverflow.com/questions/38640067/how-to-receive-client-certificate-in-an-http-request

and I check the mockwebserver,  https://github.com/square/okhttp/blob/master/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java#L493

it seems that it doesn't support client certificate yet.

ningyougang:
I found that, pass client certificate is also supported by openstack4j, in order to support keystone tokenless feature, only add pass HTTP headers, so for test case, there is no need to verify the client certificate, just check the HTTP headers is ok.

core/src/main/java/org/openstack4j/openstack/internal/BaseOpenStackService.java
```diff
@@ -112,7 +112,12 @@ protected String uri(String path, Object... params) {
         }
         RequestBuilder<R> req = HttpRequest.builder(returnType).endpointTokenProvider(ses).config(ses.getConfig())
                 .method(method).path(path);
-        return new Invocation<R>(req, serviceType, endpointFunc);
+        Map headers = ses.getHeaders();
+        if(headers != null && headers.size() > 0){
```
junoyoon: isn't it better to add space b/w if and (?

core/src/main/java/org/openstack4j/openstack/internal/BaseOpenStackService.java
```diff
@@ -112,7 +112,12 @@ protected String uri(String path, Object... params) {
...
+            return new Invocation<R>(req, serviceType, endpointFunc).headers(headers);
+        }else{
```
junoyoon: same comment as above

core/src/main/java/org/openstack4j/openstack/internal/OSClientSession.java
```diff
@@ -230,6 +232,18 @@ public CloudProvider getProvider() {
     /**
      * {@inheritDoc}
      */
+    public T headers(Map<String, ? extends Object> headers) {
+        this.headers = headers;
+        return (T) this;
```
junoyoon: what's the coding style of openstack4j? ```(T) this;``` or ```(T)this;```

core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneAuthenticationTests.java
```diff
@@ -349,6 +355,28 @@ public void reAuthentication_Test() throws Exception {
...
+     *
+     * @throws Exception
+     */
+    public void pass_headers_Test() throws Exception {
 
+        respondWith(JSON_USERS);
+
+        KeystoneToken token = new KeystoneToken();
+        Map headersMap = new HashMap();
+        headersMap.put("X-Project-Name","admin");
```
liuhongjiang: please refer to other test case, how to define variables for constant string

core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneAuthenticationTests.java
```diff
@@ -349,6 +355,28 @@ public void reAuthentication_Test() throws Exception {
...
+        headersMap.put("X-Project-Domain-Id","default");
+        Config config = Config.newConfig();
+        token.setEndpoint(authURL("/v3"));
+        OSClient.OSClientV3 osClient = OSFactory.clientFromToken(token, config).headers(headersMap);
+        osClient.identity().users().list();
+
+        RecordedRequest request = server.takeRequest();
+        System.out.println("request.getHeader(\"X-Project-Name\"):" + request.getHeader("X-Project-Name"));
```
liuhongjiang: remove this one.

core/src/main/java/org/openstack4j/openstack/internal/OSClientSession.java
ningyougang: I refer to other codes on openstack4j, it has space, (T) this, not (T)this

ningyougang:
I have rename the branch name from ningyougang:feature/support-keystone-tokenless to ningyougang:master
see: https://github.com/ContainX/openstack4j/pull/1002

PR #1001, opened 2017-04-24
SenlinNodeService: details about physical object that backs the node

sakno:
A new method in `SenlinNodeService` allows to obtain details (via `Node.getDetails()`) about physical object that backs the cluster node correctly. This can be done using `show_details` query parameter. See details about this parameter [here](https://developer.openstack.org/api-ref/clustering/?expanded=show-node-details-detail)


sakno:
Instant answer on question _Does change any behaviour to existing core services like Identity, Nova, etc?_ is no.

auhlig:
Thanks for contributing @sakno. I'll take a look during the weekend.

auhlig:
Many thanks for this @sakno

PR #1002, opened 2017-04-27
Add http headers for supporting keystone tokenless

ningyougang:
Currently, keystone supports client certificate without having to issue a
token, so it is necessary to add http headers to support this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,35 @@ OSClientV3 os = OSFactory.builderV3()
                 .scopeToProject(Identifier.byId("project id"))
                 .authenticate());
 ```
+(5) authenticate using client certificate
+```bash
+openssl pkcs12 -export -out client-certificate-keystore.p12  -inkey key.pem -in cert.pem -certfile ca.pem
+Enter Export Password:encrypt
+Verifying - Enter Export Password:encrypt
+```
+```java
+String encrypt =  "encrypt";
+KeyStore keyStore = KeyStore.getInstance("PKCS12");
+keyStore.load(new FileInputStream(new File("client-certificate-keystore.p12")), encrypt.toCharArray());
+SSLContext sslContext = SSLContexts.custom()
+        //ignore server verify
+        .loadTrustMaterial(new TrustStrategy() {
+            @Override
+            public boolean isTrusted(X509Certificate[] chain, String authType) throws CertificateException {
+                return true;
+            }
+        })
+        .loadKeyMaterial(keyStore,encrypt.toCharArray())
+        .build();
+Config config = Config.newConfig();
+config.withSSLContext(sslContext);
+OSClient.OSClientV3 osClient = OSFactory.builderV3()
+        .endpoint("https://<fqdn>:5000/v3")
+        .withConfig(config)
+        .scopeToProject(Identifier.byId("project id"))
+        //.scopeToDomain(Identifier.byId("domain id"))
+        .authenticate();
+```
 
 #### Identity Operations (Keystone) V3
 
--- a/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneTokenlessTest.java
+++ b/core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneTokenlessTest.java
@@ -1,0 +1,45 @@
+package org.openstack4j.api.identity.v3;
+
+import okhttp3.mockwebserver.RecordedRequest;
+import org.openstack4j.api.AbstractTest;
+import org.openstack4j.api.OSClient;
+import org.openstack4j.core.transport.ClientConstants;
+import org.openstack4j.model.common.Identifier;
+import org.openstack4j.openstack.OSFactory;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Tests the Identity/Keystone API version 3 Keystone tokenless
+ */
+@Test(groups = "idV3", suiteName = "Identity/V3/Keystone")
+public class KeystoneTokenlessTest extends AbstractTest {
+
+    private static final String JSON_USERS = "/identity/v3/users.json";
+    private static final String DOMAIN_ID = "default";
+
+    /**
+     * check headers whether right from request
+     *
+     * @throws Exception
+     */
+    public void pass_headers_Test() throws Exception {
+
+        respondWith(JSON_USERS);
+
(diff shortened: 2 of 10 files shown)


ningyougang:
I have renamed the branch name from ningyougang:feature/support-keystone-tokenless to ningyougang:master.
so i close this pr: https://github.com/ContainX/openstack4j/pull/998

ningyougang:
@auhlig , i have added test case for this pr, can you review and merge it?

core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneTokenlessTest.java
```diff
@@ -0,0 +1,49 @@
...
+     * check headers whether right from request
+     *
+     * @throws Exception
+     */
+    public void pass_headers_Test() throws Exception {
+
+        respondWith(JSON_USERS);
+
+        Map header = new HashMap();
+        String key = "X-Domain-Id";
```
vinodborole: Can the key be same as the example which is "X-Project-Domain-Id"?

vinodborole:
Apart from the minor comment rest LGTM

core-test/src/main/java/org/openstack4j/api/identity/v3/KeystoneTokenlessTest.java
ningyougang: * domainScope:
X-Domain-Id:If specified, its the domain scope.
* projectScope:
X-Project-Id: If specified, its the project scope.
X-Project-Domain-Id:If specified, its the domain of project scope.

here, i test it X-Domain-Id for domainScope.
please refer to:https://docs.openstack.org/developer/keystone/configure_tokenless_x509.html

vinodborole: makes sense; thanks for clarifying

auhlig:
Sorry for the delay. 
I find it slightly misleading that we use `clientFromToken()`.
(1) As stated in the other PR it would make more sense to me to create a new method `clientFromCertificate()`.

(2) Maybe the tokenless authentication could also be invoked like
```
OSClientV3 os = OSFactory.builderV3()
                .endpoint("http://<fqdn>:5000/v3")
                .certificate(cert,key)
                .scopeToProject(Identifier.byId("project id")) // or .scopeToDomain(Identifier.byId("domain id"))
                .authenticate());
```
The `scopeTo..` would be responsible for setting the required headers mentioned above.

What do you think @ningyougang, @vinodborole, @junoyoon?

ningyougang:
@auhlig @junoyoon @vinodborole 
* For suggestion one: clientFromToken()------->clientFromCertificate()
I agree with you!
* For suggestion two: Maybe the tokenless authentication could also be invoked like
For java language, it doesn't support use cert and key directly, should convert cert and key to PKCS12 format file using openssl tool.
```
openssl pkcs12 -export -out client-certificate-keystore.p12  -inkey key.pem -in cert.pem -certfile ca.pem
Enter Export Password:password
Verifying - Enter Export Password:password
```
Actully, openstack4j already support passing pkcs12 format file.
Another reason, if modified to the invoked method you gives, it is hard to implement.

* so, what about this?
  * Define a method 
```
public static OSClientV3 clientFromCertificate(String endpoint, Config config)
```

ningyougang:
@auhlig @junoyoon @vinodborole
* For @auhlig 's suggestion one:
I have added the clientFromCertifiate method

* For @auhlig 's suggestion two:
Can't use .certificate(cert,key), because java doesn't support this, java supports pkcs12 format file, so should
use openssl tool to convert cert and key to pkcs12 format file.
It is better not to use .scopeToProject(Identifier.byId("project id")) here.beacuse scopeToProject is for api for builder, not client. so this may mislead for clientFromToken api.

ningyougang:
@auhlig @junoyoon @vinodborole 
I have modified the entrypoint call method like this
```
String encrypt =  "encrypt";
KeyStore keyStore = KeyStore.getInstance("PKCS12");
keyStore.load(new FileInputStream(new File("client-certificate-keystore.p12")), encrypt.toCharArray());
SSLContext sslContext = SSLContexts.custom()
        //ignore server verify
        .loadTrustMaterial(new TrustStrategy() {
            @Override
            public boolean isTrusted(X509Certificate[] chain, String authType) throws CertificateException {
                return true;
            }
        })
        .loadKeyMaterial(keyStore,encrypt.toCharArray())
        .build();
Config config = Config.newConfig();
config.withSSLContext(sslContext);
OSClient.OSClientV3 osClient = OSFactory.builderV3()
        .endpoint("https://<fqdn>:5000/v3")
        .withConfig(config)
        .scopeToProject(Identifier.byId("project id"), Identifier.byId("domain id"))
        //.scopeToDomain(Identifier.byId("domain id"))
        .authenticate();
```

README.md
```diff
@@ -177,6 +177,35 @@ OSClientV3 os = OSFactory.builderV3()
...
+            }
+        })
+        .loadKeyMaterial(keyStore,encrypt.toCharArray())
+        .build();
+Config config = Config.newConfig();
+config.withSSLContext(sslContext);
+OSClient.OSClientV3 osClient = OSFactory.builderV3()
+        .endpoint("https://<fqdn>:5000/v3")
+        .withConfig(config)
+        .scopeToProject(Identifier.byId("project id"), Identifier.byId("domain id"))
```
auhlig: `project id` is unique across all domains. No need for `domain id`

ningyougang: hi, @auhlig , for projectScope, i have tested in my local in curl mode, it must provide project domain id or project domain name.
```
curl -v -k -s -X GET --cert /cert.pem \
     --key /key.pem \
     -H "X-Project-Name: admin" \
     https://myhost:5000/v3/roles

error as followings:
{"error": {"message": "Neither Project Domain ID nor Project Domain Name was provided.", "code": 400, "title": "Bad Request"}}
```
if i added the ` -H "X-Project-Domain-Id: default" `. it is ok.
Please refer to the check method: https://docs.openstack.org/developer/keystone/configure_tokenless_x509.html

auhlig:
You're using the `project-name`, which does require a domain scope. `project-id` is unique and works without.

ningyougang:
@auhlig ,got it, if use `project-id`, no need for domain id.
I have modified the readme.

auhlig:
Cool. Thanks for your contribution @ningyougang 💯

PR #1011, opened 2017-05-05
designate v2 service

auhlig:
This PR introduces the DNS/Designate V2 API. Currently only Zones and Recordsets operations are covered.

Note: The content of DesignateRecordsetServiceSpec.groovy and DesignateZoneServiceSpec.groovy is commented out right now, as I cannot convince betamax to actually write to the tapes. Grr! Added a couple of unit tests instead. Will take care of that next week.

/cc @gondor @vinodborole


vinodborole:
LGTM :)

PR #1015, opened 2017-05-09
Override equal and hashcode method of some Neutron classes

JackSunshine:


vinodborole:
@auhlig what do you think about this?

JackSunshine:
hi @auhlig could you take some time review this ?
Thanks

auhlig:
Sorry for the delay and many thank for this @JackSunshine.

PR #932, opened 2017-02-13
Cannot create HostAggregate

Cherry522:
I just tested the 3.0.4-SNAPSHOT version, but it still cannot create HostAggregate.
**Error messages：**
org.openstack4j.api.exceptions.ClientResponseException: Invalid input for field/attribute aggregate. Value: {u'deleted': False, u'name': u'aggregate_cy_test', u'avaiablityZone': u'uec_zone_1', u'availability_zone': u'uec_zone_1'}. Additional properties are not allowed (u'deleted', u'avaiablityZone' were unexpected)

I modified the source code to solve this problem.

PR #1017, opened 2017-05-09
cherry-picking #932

auhlig:
Cherry picking from #932. Could you take a look @vinodborole?


auhlig:
breaking changes in 2 method would trigger 3.1.0 release

vinodborole:
LGTM

PR #1020, opened 2017-05-10
Api for glance image cache, list cached

esommar:
Hi, 
This is api + unit tests for listing cached images in the glance image cache. I had some design concerns that I wrote about in: https://groups.google.com/forum/#!topic/openstack4j/HC3eA_C_iM0
I hope that this implementation is a good compromise but the disadvantage is that the empty list is returned both if the cache is empty AND if the cache is not enabled. Since a http 404 is returned if the cache is not enabled, it is not straight forward for me to understand how to separate these cases. Please have a look!

Br,
Elina


core/src/main/java/org/openstack4j/api/image/ImageService.java
```diff
@@ -1,23 +1,31 @@
...
+import java.util.Map;
+
 /**
  * OpenStack (Glance) Image based Operations
  * 
  * @author Jeremy Unruh
  */
-public interface ImageService extends RestService {
+public interface
+ImageService extends RestService {
```
vinodborole: can you please reformat this line of code?

core/src/main/java/org/openstack4j/openstack/common/CustomEpochToDateDeserializer.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Date;
+
+/**
+ * @author esommar on 5/10/2017.
+ */
+public class CustomEpochToDateDeserializer extends JsonDeserializer<Date> {
+
+    @Override
+    public Date deserialize(JsonParser jsonparser, DeserializationContext context)
+            throws IOException {
```
vinodborole: Reformat this as well

vinodborole:
@esommar thanks for your contributions, i have few minor comments there. Also, can you tell us the API output from openstack when cache is empty and when cache is not enabled?

esommar:
Thanks!
The official openstack documentation for this api is unfortunately not as good as the other documented apis, it simply says: "If the cachemanage middleware is enabled in the application pipeline, call GET /cached-images to see a JSON-serialized list of mappings that show cached images, the number of cache hits on each image, the size of the image, and the times they were last accessed." 
I have based the implementation on the curl output that I receive locally, you can find the "empty cache" and "not enabled cache" responses respectively below! Please let me know if you think I should change anything else than the formatting. Br, Elina

Curl REQ: curl -g -i --cacert "/etc/ssl/certs/CEE/ctrl-ca.crt" -X GET http://192.168.2.30:9292/v1/cached_images -H "User-Agent: python-glanceclient" -H "Content-Type: application/octet-stream" -H "X-Auth-Token: gAAAAABZFCLJKARh_l5aviAicqGmD6yQV28xVoTkRIf5ub1QI4nmj3cqolQ7I7CwwZu6tQVwk-l3Dr09UckLliowxYHlGrl4U3ht5oqUTL68P9EbVdAHvWLQCBO8RFAZWIIW2pekTdMmxYDM-H6sWkTFxWo_KNFRcPoEORup_19wnP2boC27Gk4"

RESP:
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
Content-Length: 21
X-Openstack-Request-Id: req-71bd680d-2e2f-42f5-8938-9164587c0039
Date: Thu, 11 May 2017 08:40:17 GMT

{"cached_images": []}

Curl REQ: curl -g -i --cacert "/etc/ssl/certs/CEE/ctrl-ca.crt" -X GET http://192.168.2.30:9292/v1/cached_images -H "User-Agent: python-glanceclient" -H "Content-Type: application/octet-stream" -H "X-Auth-Token: gAAAAABY92ZlJ7wyUalD-j1ZmTDG2eQ06y8MrmuNGoU5IelSXflDsXNLxagrfD8tQ5uG5yrAmKcBF42FT-qj4wnFzcZk4GLiZ6ySocvrbZlw-YkHytH5Zuys3avb9OMC_6u-RF_VF2A7InaF0R7xaFdBvtHX45yn9DmFFDvQZfjxUJkx3DWZhUU"
 
RESP:
 
HTTP/1.1 404 Not Found
Content-Length: 154
Content-Type: text/html; charset=UTF-8
X-Openstack-Request-Id: req-c99938cd-b09f-4d2f-a0f3-659d5c8bdd34
Date: Wed, 19 Apr 2017 13:30:36 GMT

vinodborole:
So with not enabled cache it will result into null List. I think that should be fine; what do you think @auhlig ?

esommar:
Hi! I have done some thinking and realized that we can in fact return null in the "cache not enabled" case using the ExecutionOptions and PropagateOnStatus.on(404). I have updated the code (+ solved some issues that I stumbled upon with the resteasy connector etc). I think that this solution is the best possible. Please have a look @vinodborole or @auhlig

vinodborole:
LGTM, @auhlig ?

core/src/main/java/org/openstack4j/model/image/CachedImage.java
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @return date when the image was last modified in the cache
+     */
+    Date getLastModified();
+
+    /**
+     *
+     * @return nr of cache hits
+     */
+    int getHits();
```
auhlig: Any reason not to use `Integer`?

esommar:
Thanks, I have pushed the change according to your comment!

auhlig:
Merged. Thanks for your contribution @esommar 👍

Issue #1012, opened 2017-05-08
deserialize error when get hypervisor's info .


dreadbird:
When I use the api" org.openstack4j.openstack.compute.internal.ext.HypervisorServiceImpl.list(HypervisorServiceImpl.java:23)"
to get the hypervisors,throw a exception:
`Can not deserialize instance of java.lang.String out of START_ARRAY token
 at [Source: {"model": ["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"], "vendor": ["GenuineIntel"], "nodes": 1, "topology": {"cores": 12, "threads": 24}}; line: 1, column: 2] (through reference chain: org.openstack4j.openstack.compute.domain.ext.HypervisorCPUInfo["model"])
com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.lang.String out of START_ARRAY token
 at [Source: {"model": ["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"], "vendor": ["GenuineIntel"], "nodes": 1, "topology": {"cores": 12, "threads": 24}}; line: 1, column: 2]`

because the model and vendor of the "cpu_info" is a Object instead of a String.
In the openstack rest api,it return as flow:

`
"cpu_info": "{"model": ["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"], "vendor": ["GenuineIntel"], "nodes": 1, "topology": {"cores": 12, "threads": 24}}",
`
the hypervisor type is :
`"hypervisor_type": "VServer"`

there is a note in the openstack api description:
`
Since version 2.28 cpu_info field is returned as a dictionary instead of string.
`

auhlig:
Thanks for investigating @dreadbird. Since you already found the solution would you like to submit a fix?

dreadbird:
Okay, my colleage will fix it and submit.

zizhongwei:
I  will fix it.

auhlig:
Fixed by @zizhongwe via #1028 .Can we close this issue?

PR #1028, opened 2017-05-23
Fix cpu info in hypervisor

zizhongwei:
Fix cpu info in hypervisor for #1012


auhlig:
Hi @zizhongwei,
Thanks for the PR.
Could you wrap some test around that? Why are you using `List<Object>`?

zizhongwei:
there is a note in the openstack api description:
```Since version 2.28 cpu_info field is returned as a dictionary instead of string.```
When I use the Hypervisor list api to get the hypervisors,it returned a JSON like this：
``` {Source: {"model": ["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"], "vendor": ["GenuineIntel"], "nodes": 1, "topology": {"cores": 12, "threads": 24}}}```
in this JSON, "model" field is ```["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"]``` and "vendor" field is ```["GenuineIntel"]```.
In my test the Jackson can reads this "model" field to ```List<Object>```
and the Jackson can also reads JSON ```"model": "Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"``` to ```List<Object>```，
so I replaced ```String``` with ```List<Object>```

auhlig:
Wouldn't​ `List<Object>` be incompatible with the jsom-string, whereas the type string would work for both?

zizhongwei:
openstack4j has used Jackson read json to java Object. Jackson can read json-string :```"model": "Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"``` to ```List<Object>```, also can read json :```"model": ["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"]``` to  ```List<Object>```; Jackson can not read json :```"model": ["Intel(R)Xeon(R)CPUE5-2620v2@2.10GHz"]``` to  ```String```  : #1012

auhlig:
I see. LGTM. Thanks @zizhongwei

Issue #980, opened 2017-03-21
neutron: port.extra_dhcp_opts redundancy parameter: optName, optValue


billy7x17:
my code is like this:
`os.networking().port().update(builder.extraDhcpOpt(NeutronExtraDhcpOptCreate.builder().optName("bootfile-name").optValue("pxelinux.0").build()).build());`

then I debug neutron log:
`2017-03-20 21:45:22.754 26898 DEBUG neutron.api.v2.base [req-9ce4cc59-7496-4ab6-87f4-b4fb119b3080 8a2b249ca4474cd9af64d51685391f2c cf97c2d4b7634bb59bd2e4f8f128b54b - - -] Request body: {u'port': {u'name': u'test', u'allowed_address_pairs': [], u'admin_state_up': True, u'extra_dhcp_opts': [{u'opt_value': u'pxelinux.0', u'optName': u'bootfile-name', u'optValue': u'pxelinux.0', u'opt_name': u'bootfile-name'}, {u'opt_value': u'10.10.127.224', u'optName': u'server-ip-address', u'optValue': u'10.10.127.224', u'opt_name': u'server-ip-address'}, {u'opt_value': u'10.10.127.224', u'optName': u'tftp_server', u'optValue': u'10.10.127.224', u'opt_name': u'tftp_server'}, {u'opt_value': u'pxelinux.0', u'optName': u'bootfile_name', u'optValue': u'pxelinux.0', u'opt_name': u'bootfile_name'}], u'device_owner': u'', u'fixed_ips': [{u'subnet_id': u'31690712-383d-45b3-86c8-c71063c15f20', u'ip_address': u'193.111.12.6'}], u'security_groups': [u'36ceb9ba-d1a0-4ed5-b688-f0cca407b5ef'], u'device_id': u''}} prepare_request_body /usr/lib/python2.7/site-packages/neutron/api/v2/base.py:684`

Including this:
{u'opt_value': u'pxelinux.0', **u'optName**': u'bootfile-name', **u'optValue**': u'pxelinux.0', u'opt_name': u'bootfile-name'}

then neutron api raise error.

I mean optName and optValue should not exist in the params.

auhlig:
Thanks for reporting.
Strange. This looks like it should work.
Maybe changing [the code](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/networking/domain/NeutronExtraDhcpOptCreate.java#L14-L15) to
```
[...]

PR #1029, opened 2017-05-23
Fix NeutronExtraDhcpOptCreate

zizhongwei:
Fix NeutronExtraDhcpOptCreate add JsonProperty to opt_value&opt_name for #980


auhlig:
Many thanks @zizhongwei

PR #1030, opened 2017-05-24
Add additional properties of the availability zone hint for the network.

taemin77:
Hi, I Added two perperties for availability zone for the network.
-  availability_zone_hints : The availability zone candidate for the network. (create and read)
-  availability_zones : 	The availability zone for the network. (read only)
* these properties is for Mitaka or above.


auhlig:
Thanks @taemin77.
Could you add some tests?

auhlig:
Thanks @taemin77

Issue #1019, opened 2017-05-10
Volume snapshot object's getName() method error,return with null. but the API actually returned the name value.


zhoutiekui:
OS4J Version:
3.0.4-snapshot

Test Code:
		VolumeSnapshot snapshotForce = os.blockStorage().snapshots().create(Builders.volumeSnapshot()
				.name(snapshotName)
				.volume(volume.getId())
				.force(true)
				.build());
		System.out.println("snapshot id:\t" + snapshotForce.getId());
		System.out.println("snapshot name:\t" + snapshotForce.getName());
		System.out.println(os.blockStorage().snapshots().get(snapshotForce.getId()).getId());
		assertEquals(snapshotName, os.blockStorage().snapshots().get(snapshotForce.getId()).getName());

Output:
snapshot id:	6980f66a-110d-4bcb-a7f9-eb8752395799
snapshot name:	null

Analyze:
the package org.openstack4j.openstack.storage.block.domain has following codes:
	@JsonProperty("display_name")
	private String name;
	@JsonProperty("display_description")
	private String description;
While the API definition the name "name" not the desplay_name. 
The same error with the description attribute.

auhlig:
Thanks for investigating @zhoutiekui. Feel encouraged to contribute a fix for that :)

cbirajdar:
Hi @auhlig! I have also noticed this as an issue with CinderVolume - getName() and getDescription(). For block storage API V1, field names were `display_name` and `display_description`. Since V2, it's `name` and `description`.

```
@JsonRootName("volume")
public class CinderVolume implements Volume {
	@JsonProperty("display_name")
	private String name;
	@JsonProperty("display_description")
        private String description;
```

Is version 1 still supported for managing cinder volumes? I can simply rename these properties in CinderVolume and update V1/V2 related unit tests which would otherwise fail.

Thank you!

auhlig:
This was fixed by now, wasn't it? Could you try the latest snapshot?

cbirajdar:
For block storage API V2, it was fixed for [CinderVolumeSnapshot](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/storage/block/domain/CinderVolumeSnapshot.java#L28) but still an issue with [CinderVolume](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/storage/block/domain/CinderVolume.java#L30).

PR #1031, opened 2017-05-26
Fix cinder volume snapshot

zizhongwei:
change @JsonProperty("name") for name & change @JsonProperty("description") for description;
for #1019


auhlig:
Was it always just name and description or did that change recently @zizhongwei ?

zizhongwei:
it was just name and description since Block Storage API V2 ： 
image

auhlig:
Thanks for explaining.

PR #1034, opened 2017-05-30
Added missing fields in Magnum bay

sohaninfo:
Some fields were missing in Magnum bay model. This check-in adds those missing fields.

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/magnum/MagnumBay.java
+++ b/core/src/main/java/org/openstack4j/openstack/magnum/MagnumBay.java
@@ -9,7 +9,7 @@ import org.openstack4j.openstack.common.ListResult;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.base.Objects;
+import com.google.common.base.MoreObjects;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class MagnumBay implements Bay {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/openstack4j/openstack/magnum/MagnumBay.java
```diff
@@ -7,172 +7,273 @@
 import org.openstack4j.openstack.common.GenericLink;
 import org.openstack4j.openstack.common.ListResult;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.base.MoreObjects;
```
vinodborole: Any reason why remove this? this was added for the latest version of guava lib

auhlig: I initially got the impression Objects was deprecated. Seems that's not the case. What's the difference?

sodre90: https://google.github.io/guava/releases/20.0/api/docs/com/google/common/base/Objects.ToStringHelper.html

Deprecated. 
Use MoreObjects.ToStringHelper instead. This class is scheduled for removal in Guava 21.0.

Objects.ToStringHelper is depreated. I changed it to MoreObjects because we use guava 21 in our project. In guava 21 Objects.ToStringHelper has been removed, so we got NoSuchMethodError in runtime.

Anyway I updated only for guava 20, because guava 21 works with only jdk8. guava 22 supports jdk7 again with 22.0-android version.

sohaninfo: Since Objects.ToStringHelper is depreated, I have changed it to MoreObjects.ToStringHelper.

auhlig: Ah. Obviously you found the correct docs. Sorry for the confusion and thanks @sodre90.

auhlig:
LGTM. Is this (unit-) tested? Else could you add a small check @sohaninfo?

sohaninfo:
@auhlig yes, I have done unit-testing. It works fine.

auhlig:
I see. Thanks for this @sohaninfo

PR #1039, opened 2017-06-05
Mistral: implement endpoints for action definitions and workbook definitions

rakhmerov:
Added two more endpoints for Mistral:
* Action definitions
* Workbook definitions

### DIFF
--- a/core-test/src/main/java/org/openstack4j/api/workflow/ActionDefinitionTest.java
+++ b/core-test/src/main/java/org/openstack4j/api/workflow/ActionDefinitionTest.java
@@ -1,0 +1,129 @@
+package org.openstack4j.api.workflow;
+
+import org.openstack4j.model.common.ActionResponse;
+import org.openstack4j.model.workflow.ActionDefinition;
+import org.openstack4j.model.workflow.Scope;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+import static org.testng.Assert.*;
+
+/**
+ * Test cases for {@link ActionDefinitionService}.
+ * 
+ * @author Renat Akhmerov
+ */
+@Test(suiteName="ActionDefinitions")
+public class ActionDefinitionTest extends WorkflowBaseTest {
+
+    private static final String JSON_ACTION_DEF = "/workflow/action_def.json";
+    private static final String JSON_ACTION_DEFS = "/workflow/action_defs.json";
+    private static final String NEW_ACTION = "/workflow/new_action.yaml";
+    private static final String JSON_ACTION_DEF_CREATE = "/workflow/action_def_create.json";
+
+    private ActionDefinitionService service;
+
+    @BeforeTest
+    public void setUp() {
+        this.service = osv3().workflow().actionDefinitions();
+    }
+
+    @Test
+    public void listActionDefinitions() throws Exception {
+        respondWith(JSON_ACTION_DEFS);
+        
+        List<? extends ActionDefinition> actionDefs = service.list();
+
+        assertEquals(actionDefs.size(), 2);
+
+        // Check first action definition.
+        ActionDefinition actionDef = actionDefs.get(0);
+
+        assertNotNull(actionDef);
+        assertIsUUID(actionDef.getId());
+        assertNotEmptyString(actionDef.getProjectId());
+        assertEquals(actionDef.getName(), "concat");
+        assertNotNull(actionDef.getTags());
+        assertEquals(actionDef.getTags().get(0), "test");
+        assertEquals(actionDef.getTags().get(1), "custom");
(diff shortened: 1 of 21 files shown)


rakhmerov:
@vinodborole, @auhlig: one more PR

core-test/src/main/java/org/openstack4j/api/workflow/WorkbookDefinitionTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertNull(wbDef.getUpdatedAt());
+        assertNotEmptyString(wbDef.getDefinition());
+        assertTrue(wbDef.getDefinition().contains("with-items: i in <% range(0, 40) %>"));
+    }
+
+    @Test
+    public void deleteWorkbookDefinition() throws Exception {
+        respondWith(204); // No content.
+
+        HttpResponse resp = service.delete("my_wb");
```
auhlig: Use `ActionResponse` instead of HttpResponse

core-test/src/main/java/org/openstack4j/api/workflow/ActionDefinitionTest.java
```diff
@@ -0,0 +1,128 @@
...
+        assertNotEmptyString(actionDef.getDefinition());
+        assertTrue(actionDef.getDefinition().contains("base: std.echo"));
+        assertEquals(actionDef.getInput(), "s1, s2");
+    }
+
+    @Test
+    public void deleteActionDefinition() throws Exception {
+        respondWith(204); // No content.
+
+        HttpResponse resp = service.delete("concat");
```
auhlig: Use `ActionResponse` instead of HttpResponse

core/src/main/java/org/openstack4j/openstack/workflow/internal/WorkbookDefinitionServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+        return invocation.entity(new InputStreamPayload(wfText)).execute();
+    }
+
+    @Override
+    public WorkbookDefinition get(String identifier) {
+        return get(MistralWorkbookDefinition.class, uri("/workbooks/%s", identifier)).execute();
+    }
+
+    @Override
+    public HttpResponse delete(String identifier) {
```
auhlig: `ActionResponse`

rakhmerov: Ok, let me recall.. I used HTTPResponse because ActionResponse masks a real HTTP code. No matter what I configured with "respondWith()" (e.g. 204) in tests it always returned 200. This is where it happens: https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/core/transport/HttpEntityHandler.java#L63
Above that line there's a condition "if (response.getStatus() >= 400) {". So id status is less than 400 then it always returns "ActionResponse.actionSuccess()" which is 200.

On the other hand, precise 204 code returned by Mistral endpoint for "delete" may not be so important in the Java client.

What do you think? Maybe I don't fully understand semantics of ActionResponse.

Any help would be useful.

Thanks

rakhmerov:
@auhlig Please comment once you have time so that I could continue. Thanks

rakhmerov:
@vinodborole Maybe you can help too.

rakhmerov:
@auhlig, @vinodborole: How about adding one more method ActionResponse.actionSuccess(int code) so that we could specify a precise http code? Existing method ActionResponse.actionSuccess() will just default to 200. And is success will be checking a range >=200 && <300 ?

vinodborole:
@rakhmerov  +1 for adding a new method

rakhmerov:
@vinodborole, @auhlig: Ok, then I'll just change to ActionResponse for now which will contain 200 code and then add that new method in a separate patch so that it's easier to track.

rakhmerov:
@vinodborole, @auhlig: done

auhlig:
Why is it necessary to distinguish between 200 and 204? 
We just want to know wether the Action was successful (as indicated by 200/204) or if it wasn't.

vinodborole:
LGTM

rakhmerov:
@auhlig Well, just to be aligned with HTTP protocol and keep the exact code that the server replied with. Even if we look at our tests we'll see that there's the method respondWith(int code) that takes code but it's pointless to test with codes other than 200, because our infrastructure can't pass them precisely.

I made changes to keep the exact code locally, they are really trivial. I was going to push them along with some other changes I made (not 100% ready yet). I'm not insisting though. If you think it's not needed I'll roll back these changes.

Anyway, hopefully this PR is ok now. Waiting for your decision.

Thanks

auhlig:
This PR looks good. Thanks @rakhmerov.

rakhmerov:
@auhlig Ok, thanks!

PR #1046, opened 2017-06-14
Renamed CinderVolume properties to be compatible with block storage API v2, fixed broken VolumeTests

### LINKED ISSUE #1019: Volume snapshot object's getName() method error,return with null. but the API actually returned the name value.

cbirajdar:
Fixes #1019. For v2/v3, getName() and getDescription() should return the correct values. For v1, it will return null as API fields since v2 are updated to `name` and `description`.

### DIFF
--- a/core-test/src/main/java/org/openstack4j/api/storage/VolumeTests.java
+++ b/core-test/src/main/java/org/openstack4j/api/storage/VolumeTests.java
@@ -63,6 +63,8 @@ public class VolumeTests extends AbstractTest {
         assertTrue(getRequest.getPath().matches("/v[12]/\\p{XDigit}*/volumes/8a9287b7-4f4d-4213-8d75-63470f19f27c"));
         
         assertEquals(volume.getId(), "8a9287b7-4f4d-4213-8d75-63470f19f27c");
+        assertEquals(volume.getDisplayName(), "vol-test");
+        assertEquals(volume.getDisplayDescription(), "a description");
         assertNotNull(volume.getCreated());
         assertEquals(volume.getZone(), "nova");
         assertEquals(volume.getSize(), 100);
--- a/core-test/src/main/java/org/openstack4j/api/storage/VolumeTests.java
+++ b/core-test/src/main/java/org/openstack4j/api/storage/VolumeTests.java
@@ -63,8 +63,6 @@ public class VolumeTests extends AbstractTest {
         assertTrue(getRequest.getPath().matches("/v[12]/\\p{XDigit}*/volumes/8a9287b7-4f4d-4213-8d75-63470f19f27c"));
         
         assertEquals(volume.getId(), "8a9287b7-4f4d-4213-8d75-63470f19f27c");
-        assertEquals(volume.getName(), "vol-test");
-        assertEquals(volume.getDescription(), "a description");
         assertNotNull(volume.getCreated());
         assertEquals(volume.getZone(), "nova");
         assertEquals(volume.getSize(), 100);
@@ -91,7 +89,6 @@ public class VolumeTests extends AbstractTest {
     
     @SuppressWarnings("unchecked")
     @Test
-    @SkipTest(connector = ".*", issue = 395, description = "Volume attribute not recognized when using cinder v2 api")
     public void getVolumeV2() throws Exception {
         // Check get volume
         respondWith("/storage/v2/volume.json");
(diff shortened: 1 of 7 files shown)


core-test/src/main/java/org/openstack4j/api/storage/VolumeTests.java
```diff
@@ -63,8 +63,6 @@ public void getVolumeV1() throws Exception {
         assertTrue(getRequest.getPath().matches("/v[12]/\\p{XDigit}*/volumes/8a9287b7-4f4d-4213-8d75-63470f19f27c"));
         
         assertEquals(volume.getId(), "8a9287b7-4f4d-4213-8d75-63470f19f27c");
-        assertEquals(volume.getName(), "vol-test");
```
vinodborole: any reason you have to remove these, the methods are the same i guess

cbirajdar: Hi @vinodborole!

As mentioned above - for block storage API v1, volume.getName() and volume.getDescription() now return null values (due to json property change `display_name->name` and `display_description->description`) and these assertions would fail. 

Similar change was applied to CinderVolumeSnapshot.java in #1031. Nevertheless, json property change fixes the response for API v2/v3. Removed the skip from getVolumeV2() which does assert for name and description.

vinodborole: @auhlig What do you think?

cbirajdar:
@auhlig When you get a chance, can you review and merge this? Thanks!

core-test/src/main/java/org/openstack4j/api/storage/VolumeTests.java
auhlig: While thinking about it: How about keeping both properties `display_name` and `name`? Maybe we should have done the same in the other class?

cbirajdar: Done. Fixed both classes, added `display_name` and `display_description` to maintain backwards compatibility with deprecated cinder v1 API.

auhlig:
Thanks @cbirajdar. LGTM. Happy with this @vinodborole?

vinodborole:
LGTM

pdube:
How do you configure your the client to use V2/V3? This fix actually broke V1 volume creation. The "name" field is no longer serialized to "display_name" so all created volumes have an empty name field

cbirajdar:
@pdube For V2/V3 before this fix, getName() and getDescription() was returning null due to the breaking changes introduced from V1 -> V2.  Openstack4j relies on the service catalog to apply default service endpoints. You can implement the EndpointURLResolver interface and use different version instead as mentioned [here](http://www.openstack4j.com/learn/getting-started/#endpoint-url-resolver).

```
public class CustomEndpointUrlResolver implements EndpointURLResolver {

    @Override
    public String findURLV2(URLResolverParams params) {
        // Identity V2
        return params.token.getEndpoint();
    }

    @Override
    public String findURLV3(URLResolverParams params) {
        // Identity V3
        if (params.type.equals(ServiceType.BLOCK_STORAGE)) {
            return BLOCK_STORAGE_V2_OR_V3_URL;
        }
}
```

V1 is deprecated as per openstack docs but in order to fix the volume creation for V1, just need to update the VolumeBuilder and set these deprecated fields to serialize into correct json attributes.

pdube:
Thanks for the quick response, I will test with the EndpointURLResolver. Since V1 is deprecated, does it make sense to change the default resolver to use cinder V3?

cbirajdar:
Default endpoints and versions are resolved based on the openstack release version and configured endpoints in the service catalog. Can you check the service catalog for your OS environment and make sure if it points to the correct cinder version (or the latest available)?

Also, I fixed the cinder V1 volume creation so that name and description will persist and won't be empty. Pull request #1058.

PR #1056, opened 2017-07-01
Added a new method createSnapshot() to invoke create image action API with metadata properties

cbirajdar:
Openstack API to [create image](https://developer.openstack.org/api-ref/compute/?expanded=create-image-createimage-action-detail#create-image-createimage-action) allows passing optional metadata properties. Overloaded `createSnapshot()` to handle additional metadata and maintain backwards compatibility.


vinodborole:
@cbirajdar thanks for this, it would really be good if we have some tests around this new method. Also, what if metadata is not null but empty; whats the behaviour of the API call?

cbirajdar:
Added unit tests for the new method. @vinodborole @auhlig Please review and merge when you get a chance. Thanks!

vinodborole:
LGTM, thanks @cbirajdar 👍

cbirajdar:
Thanks @vinodborole! Can this be merged now?

PR #1058, opened 2017-07-07
Update builder to set old deprecated fields to support volume creation (preserving name and description) for cinder API V1

cbirajdar:
Original fix in #1046 introduced a bug with V1 volume creation where new volume would result with no name or description. This is a NO-IMPACT fix to set extra fields which will be ignored for cinder API V2/V3.


vinodborole:
LGTM

PR #1061, opened 2017-07-12
Allow value to be object type in ServiceConfig

ekasitk:
This PR allows values in ServiceConfig to be an integer or any other types than just a string. So the generated JSON is correctly formatted according to the type of value.

### DIFF
--- a/core-test/src/main/java/org/openstack4j/api/AbstractTest.java
+++ b/core-test/src/main/java/org/openstack4j/api/AbstractTest.java
@@ -42,6 +42,7 @@ public abstract class AbstractTest {
         BLOCK_STORAGE(8776),
         METERING(8087),
         TELEMETRY(8087),
+        SAHARA(8386),
         SHARE(8786),
         OBJECT_STORAGE(8800),
         BARBICAN(9311),
--- a/core-test/src/main/java/org/openstack4j/api/sahara/ClusterTests.java
+++ b/core-test/src/main/java/org/openstack4j/api/sahara/ClusterTests.java
@@ -1,0 +1,95 @@
+package org.openstack4j.api.sahara;
+
+import org.openstack4j.model.sahara.Cluster;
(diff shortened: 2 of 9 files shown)


vinodborole:
@ekasitk can you please share the use case for this change request?

ekasitk:
When setting service parameters, values can be string or number.

      ServiceConfig hdfsConf = Builders.serviceConfig()
                                  .set("dfs.namenode.logging.level","info")
                                  .set("dfs.replication",1);
     Cluster cluster = Builders.cluster().name("test")
                            .addServiceConfig("HDFS",hdfsConf);

This PR permits ServiceConfig.set() to accept number and  generate the json like:
{
  "name" : "test",
  "cluster_configs" : {
    "HDFS" : {
      "dfs.namenode.logging.level" : "info",
      "dfs.replication" : 1
    }
  },
  "hadoop_version" : "1.6.2"
}

Otherwise, all values will always be treated as string, i.e. "dfs.replication" : "1", which is not  complied well to the API.  Ref: https://developer.openstack.org/api-ref/data-processing/?expanded=show-details-of-a-cluster-detail.

auhlig:
Makes sense. Or not @vinodborole?

vinodborole:
I agree @auhlig as per the openstack doc this change request makes sense. A test for a more complex config will be necessary in order to be sure nothing else breaks. @ekasitk is it possible for you to add this test case along with this PR?

ekasitk:
@vinodborole  @auhlig  
I can add some test cases but it will take a while.

vinodborole:
@ekasitk sure not a problem

ekasitk:
@vinodborole @auhlig 
I've added some test cases for this PR. Please kindly check.

vinodborole:
LGTM @auhlig what do you think?

auhlig:
LGTM. Many thanks @ekasitk

PR #1062, opened 2017-07-14
Fix ETag header to match documented name: 'Etag'

### LINKED ISSUE #994: swiftObject.getETag() is null on after calling swiftObject.download()
However, the eTag is available in the header

```
val download = obj.download()
val eTag = download.getHttpResponse.header("Etag")
```

anowell:
I was getting NULL out of the `getETag` method (using 3.0.3 with httpclient, targeting Rackspace). Per the [Swift API docs](https://developer.openstack.org/api-ref/object-storage/?expanded=get-object-content-and-metadata-detail), the header is `Etag`, not `ETag`, and I've confirmed with a custom build of OpenStack4J that this case change fixes the issue for me.

I believe this would resolve #994


auhlig:
Cool. Many thanks @anowell. Will merge after CI is green.

Issue #1068, opened 2017-07-20
Updating Firewall error  v2


scsunchen:
>  <dependency>
		  <groupId>org.pacesys.openstack4j.connectors</groupId>
		  <artifactId>openstack4j-httpclient</artifactId>
		  <version>3.0.4</version>
		

> </dependency>

**Unrecognized attribute(s) 'policy', status=400, status-code=BAD_REQUEST**

> 16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "{[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "  "firewall" : {[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "    "shared" : false,[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "    "policy" : "036ab1f1-271e-4db1-8d04-e4ed226924c8",[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "    "admin_state_up" : false,[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "    "firewall_policy_id" : "036ab1f1-271e-4db1-8d04-e4ed226924c8"[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "  }[\r][\n]"
16:05:16.505 [main] DEBUG org.apache.http.wire - http-outgoing-1 >> "}"

Cannot update

scsunchen:
NeutronFirewallUpdate is not a  @JsonIgnore ?
      @JsonIgnore
	@Override
	public String getPolicy() {
		return policyId;
	}

auhlig:
Could be a bug. Would you like to contribute and submit a fix @scsunchen ?

auhlig:
#1074 was merged. Closing issue. Please reopen if necessary.

PR #1074, opened 2017-07-27
fix  Firewall policy param cannot be updated

scsunchen:
fix  Firewall policy param cannot be updated #1068

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/networking/domain/ext/NeutronFirewallUpdate.java
+++ b/core/src/main/java/org/openstack4j/openstack/networking/domain/ext/NeutronFirewallUpdate.java
@@ -3,6 +3,7 @@ package org.openstack4j.openstack.networking.domain.ext;
 import org.openstack4j.model.network.ext.FirewallUpdate;
 import org.openstack4j.model.network.ext.builder.FirewallUpdateBuilder;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonRootName;
@@ -74,7 +75,7 @@ public class NeutronFirewallUpdate implements FirewallUpdate {
 	public Boolean isShared() {
 		return shared != null && shared;
 	}
-
+	@JsonIgnore
 	@Override
 	public String getPolicy() {
 		return policyId;


auhlig:
Hi @scsunchen, 
I'm very sorry this PR didn't get attention so far. 
I was looking at https://developer.openstack.org/api-ref/networking/v2/index.html?expanded=create-firewall-group-detail,id284-detail#id284 and didn't see the `firewall_policy_id` attribute. Looks like that was a thing in v1. 
In v2 this was replaced by `egress_firewall_policy_id` and `ingress_firewall_policy_id`.
AFAIK, OpenStack4j currently supports only networking v1 and v2 only if there was no change, but I'm not to familiar with this part. Maybe you are?

scsunchen:
@auhlig that is in v1.   It's a bug .

auhlig:
Both v1 and v2 point to /v2/-paths? Probably the docs are incorrect on the path here.
In v1 I can still see`firewall_policy_id` as a valid parameter in both request and response of the update method https://developer.openstack.org/api-ref/networking/v2/index.html?expanded=create-firewall-group-detail,id284-detail,update-firewall-rule-detail#update-firewall-rule. Which version of the networking version are you using @scsunchen?

scsunchen:
@auhlig  
when I use `os.networking().firewalls().firewall() .update()`   。`ClientResponseException{message=Unrecognized attribute(s) 'policy', status=400, status-code=BAD_REQUEST}`
`firewall_policy_id` is a valid. but `policy `is a invalid 
please look at the NeutronFirewallUpdate.java . 79 line.


{
"firewall" : {
 "shared" : false,
 "**policy**" : "036ab1f1-271e-4db1-8d04-e4ed226924c8",
 "admin_state_up" : true,
 "firewall_policy_id" : "036ab1f1-271e-4db1-8d04-e4ed226924c8"
 }
}
It has not been validated

[https://developer.openstack.org/api-ref/networking/v2/?expanded=update-firewall-detail#extensions](url)

auhlig:
What do you think on this @vinodborole?

vinodborole:
@auhlig @scsunchen After referring to the documentation for v2 i believe the parameter "policy" does not exists. So it makes sense to ignore it.

auhlig:
Since we're currently not distinguishing between v1 and v2. There just one networking service for both we're lucky they don't seem to differ much. I was concerned, as I thought merging this could break things for v1, where the policy attribute exists. Makes sense @vinodborole @scsunchen ?

scsunchen:
@vinodborole I agree with you。But look at the file : https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/networking/domain/ext/NeutronFirewallUpdate.java   in 78 line number . Where's @JsonIgnore?  
why not have a try it ?

and  look https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/networking/domain/ext/NeutronFirewall.java
in  135 line .

scsunchen:
如果还不能证明我是对的。那么我放弃提交这个更新！

auhlig:
What do you think @vinodborole?

vinodborole:
@auhlig I think it makes perfect sense to ignore the "policy" property in neutronFirewall Update as the get firewall has ignored it.

auhlig:
Thanks @scsunchen!

Issue #1044, opened 2017-06-13
Wrong return type of disable/enable service actions?


baseman79:
Hi,

I saw this one:

[ServicesService.java](https://github.com/ContainX/openstack4j/blob/ac15d6cbc3a4c2d927c91ed3691a9c60262f9b45/core/src/main/java/org/openstack4j/api/compute/ext/ServicesService.java)

    /**
     * Enables a compute services.
     *
     * @param binary the name of the service binary that you want to enable
     * @param host the host name of the service that you want to enable
     * @return the status of the enabled service
     * 
     * @author Wang Ting/王婷
     */
    Service.Status enableService(String binary, String host);

    /**
     * Disables a service.
     *
     * @param binary the name of the service binary that you want to disable
     * @param host the host name of the service that you want to disable
     * @return the status of the disabled service
     * 
     * @author Wang Ting/王婷
     */
    Service.Status disableService(String binary, String host);

The response of the above calls are different, according to [Enable Scheduling For A Compute Service](https://developer.openstack.org/api-ref/compute/?expanded=disable-scheduling-for-a-compute-service-detail,log-disabled-compute-service-information-detail,update-forced-down-detail,enable-scheduling-for-a-compute-service-detail#enable-scheduling-for-a-compute-service) and [Disable Scheduling For A Compute Service](https://developer.openstack.org/api-ref/compute/?expanded=update-forced-down-detail#disable-scheduling-for-a-compute-service), does the return type suppose to be [ExtService](https://github.com/ContainX/openstack4j/blob/d9a6417f1c33c8901cec5f3fc7b3f6caa3458fd3/core/src/main/java/org/openstack4j/openstack/compute/domain/ext/ExtService.java)?

baseman79:
So here is a proof, using 3.0.5-snapshot build against a Newton OpenStack, I'm getting an exception, and it clearly returns a service not the status. I will pull a fix shortly:

```
2017-08-05 21:34:03.434 [http-nio-127.0.0.1-8095-exec-4] INFO  os - 1 * Sending client request on thread http-nio-127.0.0.1-8095-exec-4
1 > PUT http://somehost:8774/v2/4012..../os-services/disable
1 > Accept: application/json
1 > Content-Type: application/json
1 > User-Agent: OpenStack4j / OpenStack Client
1 > X-Auth-Token: gAAAAABZ....
{
  "binary" : "nova-compute",
  "host" : "host1"
}

[...]

PR #1076, opened 2017-08-06
Wrong return type of disable/enable service actions

baseman79:
https://github.com/ContainX/openstack4j/issues/1044


vinodborole:
@baseman79 I don't see a reason why the return type should not get parsed with Service class. It does satisfies all the parameters; can you share any exceptions or errors you get?

https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/model/compute/ext/Service.java

vinodborole:
@baseman79 I don't there would be need for this PR anymore; until there is something we are trying to fix here from the older code.

baseman79:
@vinodborole 
` I don't see a reason why the return type should not get parsed with Service class. It does satisfies all the parameters; can you share any exceptions or errors you get?`

sorry, are you saying the old code with return type `Sevice.Status` will work fine?

vinodborole:
@baseman79   Yes it should

baseman79:
@vinodborole,

here is the return from OS:
```2017-08-08 10:00:05.148 [http-nio-127.0.0.1-8095-exec-5] INFO  os - 2 * Client response received on thread http-nio-127.0.0.1-8095-exec-5
2 < 200
2 < Connection: keep-alive
2 < Content-Length: 81
2 < Content-Type: application/json
2 < Date: Tue, 08 Aug 2017 14:00:10 GMT
2 < X-Compute-Request-Id: req-eacf5035-08c9-43dc-9b5c-74d5c88dbae0
{"service": {"status": "enabled", "binary": "nova-compute", "host": "somehost"}}
```

and the value I got back by using the old code

```
s	Service$Status  (id=81)	
	name	"UNRECOGNIZED" (id=87)	
	ordinal	2
```

baseman79:
Hi @vinodborole, 

Just to be clear, I made a very simple call to disable a host with the current master code (return type is ```Service.Status```)
```
Service.Status s = oc.compute().services().disableService("nova-compute", hostName);
```
OS reponsed
```
2 < 200
2 < Connection: keep-alive
2 < Content-Length: 81
2 < Content-Type: application/json
2 < Date: Tue, 08 Aug 2017 14:00:10 GMT
2 < X-Compute-Request-Id: req-eacf5035-08c9-43dc-9b5c-74d5c88dbae0
{"service": {"status": "enabled", "binary": "nova-compute", "host": "somehost"}}
```
And I verified the host is disabled, but the return value I got for ```s```  is ```UNRECOGNIZED```

Do you still think this pr is not needed?

Thanks,

vinodborole:
@baseman79 thank you for debuging this 

@auhlig what do you think?

auhlig:
We're talking https://developer.openstack.org/api-ref/compute/#disable-scheduling-for-a-compute-service and https://developer.openstack.org/api-ref/compute/#enable-scheduling-for-a-compute-service, right? Both return a service, so this PR makes perfect sense.
It also looks like this is currently not tested, so I'd like to ask if you could add a unit test.

baseman79:
@auhlig that's correct. I will add an unit test, thx

baseman79:
@auhlig @vinodborole anyone?

auhlig:
Sorry for the delay. I'll take a look in the evening.

baseman79:
no worries, thanks @auhlig

auhlig:
Thanks for contributing @baseman79

PR #1077, opened 2017-08-08
Allow identity V3 to lookup service by name

ekasitk:
According to the in identity V2, the service is lookup by name or by type (see [ref](https://github.com/ContainX/openstack4j/blob/d9a6417f1c33c8901cec5f3fc7b3f6caa3458fd3/core/src/main/java/org/openstack4j/openstack/identity/v2/domain/KeystoneAccess.java#L237)). However, Identity V3 lookup only by type. This PR intends to bring back the ability to lookup service by name.


auhlig:
LGTM. Thanks @ekasitk

PR #1078, opened 2017-08-10
Fix in ServiceTypeEnum

rafael-bertoli:
Fix method forName to return correct type when attribute name contains version number like "volumev3"


auhlig:
Although this looks good to me, could you add a small unit test for this?

auhlig:
Cool. Many thanks @rafael-bertoli

PR #1079, opened 2017-08-16
Support multiattach property in volume

taemin77:
Added multiattach property when creating volume.
multiattach (Optional) : To enable this volume to attach to more than
one server, set this value to true. Default is false.


auhlig:
LGTM. Many thanks @taemin77.

PR #1080, opened 2017-08-18
Feature/barbican secret

rschollmeyer:
Add barbican secrets implementation inspired by the Openstack Secrets API.


core/src/main/java/org/openstack4j/openstack/barbican/domain/BarbicanSecret.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.Date;
+import java.util.List;
+
+/**
+ * Created by reneschollmeyer on 02.08.17.
+ */
+public class BarbicanSecret implements Secret {
+
+    private String name;
+    @JsonProperty("algorithm")
```
auhlig: Nit: `@JsonProperty("<jsonName>")` is optional, if property name is equal in json and java

core/src/main/java/org/openstack4j/api/barbican/SecretService.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by reneschollmeyer on 02.08.17.
+ */
+public interface SecretService {
+
+    List<? extends Secret> list(final Map<String, String> filteringParams);
```
auhlig: Learning question: Why make this `final`?

PR #1081, opened 2017-08-25
Added API for Octavia

Grim-lock:
octavia API has some difference from neutron lbaas API, so create a separate directory, and I will add L7 API later.


auhlig:
That was a lot of work. Many thanks @Grim-lock!
Although this looks very promising could you add tests?

Grim-lock:
@auhlig I would like to do it, but I can not know how much time is spent, maybe 2 weeks or 3 weeks.

Grim-lock:
@auhlig finished the test case

auhlig:
Perfect @Grim-lock! Many thank. Good to merge from my side.
What do you think @vinodborole ?

vinodborole:
@Grim-lock  many thanks. LGTM

PR #1086, opened 2017-09-14
Support Availability zones API in Neutron.

taemin77:
Hi, I added api for neutron.

Reference Url : https://docs.openstack.org/mitaka/networking-guide/config-az.html

Availability zones  :   An availability zone groups network nodes that run services like DHCP, L3, FW, and others. It is defined as an agent’s attribute on the network node. This allows users to associate an availability zone with their resources so that the resources get high availability.


auhlig:
Great. Thanks @taemin77

PR #1087, opened 2017-09-18
fix(secret): fix wrong representation of Secret values & add payload to Secrets

rschollmeyer:
Some minor fixes to match the Barbican API.


rschollmeyer:
Made a small mistake & will open a new pull request.

PR #1088, opened 2017-09-18
fix(secret): fix wrong representation of Secret values & add payload

rschollmeyer:
Some minor fixes to match the Barbican API.

### DIFF
--- a/core-test/src/main/java/org/openstack4j/api/barbican/SecretTests.java
+++ b/core-test/src/main/java/org/openstack4j/api/barbican/SecretTests.java
@@ -8,6 +8,7 @@ import org.openstack4j.model.common.ActionResponse;
 import org.testng.annotations.Test;
 
 import java.io.IOException;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
@@ -25,7 +26,7 @@ public class SecretTests extends AbstractTest {
 
     private final String secretId = "520405bc-c7c5-41ea-97ad-6c67a8d41a9e";
     private final String secretName = "test_secret";
-    private final String expiration = "2015-12-28T19:14:44.180394";
+    private final String expiration = "Mon Dec 28 20:17:44 CET 2015";
     private final String content_type = "application/octet-stream";
 
     public void testListSecretsByName() throws IOException {
@@ -47,7 +48,7 @@ public class SecretTests extends AbstractTest {
         Secret secret = osv3().barbican().secrets().get(secretId);
         assertNotNull(secret);
         assertNotNull(secret.getName());
-        assertEquals(secret.getExpiration(), expiration);
+        assertEquals(secret.getExpiration().toString(), expiration);
         assertTrue(!secret.getContentTypes().isEmpty());
         assertEquals(secret.getContentTypes().get("default"), content_type);
     }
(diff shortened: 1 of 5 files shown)


auhlig:
Many thanks @rschollmeyer. Could you extend the tests accordingly?

rschollmeyer:
Oh right. I'm on it!

rschollmeyer:
Actually there was not much to extend since the changes were mainly changes in representation.

core/src/main/java/org/openstack4j/openstack/barbican/domain/BarbicanSecret.java
```diff
@@ -18,16 +19,17 @@
...
-    private List<String> contentTypes;
+    private Integer bitLength;
+    @JsonProperty("content_types")
+    private Map<String, String> contentTypes;
     @JsonProperty("created")
-    private Date createTime;
+    private String createTime;
     @JsonProperty("updated")
-    private Date updateTime;
+    private String updateTime;
```
auhlig: Why Date -> String?

rschollmeyer: Because it is listed as a String in the Barbican API.

See [here](https://docs.openstack.org/barbican/latest/api/reference/secrets.html#get-v1-secrets)

rschollmeyer:
I thought of that too, but the Secret object that is returned when creating seems to only contain the reference.

core/src/main/java/org/openstack4j/openstack/barbican/domain/BarbicanSecret.java
auhlig: I was asking, because changing the type is a breaking change and it might be more convenient for our Java client to have a Date here instead of a String. But since you're the author I would leave that up to you.
Besides this, this PR is ready to merge from my side.

rschollmeyer: Well, i get your point & Date would probably be better. I'll change it back

rschollmeyer:
Done the changes. Note that the output of Dates when calling GET for a secret is now "Mon Dec 28 19:17:44 UTC 2015" instead of "2015-12-28T19:14:44.180394" (as specified in the barbican API)

core-test/src/main/java/org/openstack4j/api/barbican/SecretTests.java
```diff
@@ -47,14 +48,25 @@ public void testGetSecret() throws IOException {
         Secret secret = osv3().barbican().secrets().get(secretId);
         assertNotNull(secret);
         assertNotNull(secret.getName());
+        assertEquals(secret.getExpiration().toString(), expiration);
```
auhlig: Instead of comparing the strings, I'd suggest to compare the Date types. Make sense? I promise that's the last comment @rschollmeyer.

rschollmeyer: Sure i can do that. Don't worry, i am thankful for every good critique!

rschollmeyer: To be sure: do you mean to compare the date formats?

auhlig: Compare `assertEquals(secret.getExpiration(), expiration);` where expiration is a `Date` object

auhlig:
Perfect. Thanks for all the work @rschollmeyer  💯

PR #1092, opened 2017-09-27
Switch to standalone snakeyaml

olivergondza:
While experimenting with updating jackson version I noticed, `com.fasterxml.jackson.dataformat.yaml.snakeyaml.*` is [no longer exposed](https://github.com/FasterXML/jackson-dataformat-yaml/commit/c69224917fd235732e51c8327f23af7f7038108f) so I am switching to proper dependency which should hopefully resolve the OSGi limitation as well.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -74,7 +74,6 @@
 						</Private-Package>
 						<Import-Package>
 							!org.openstack*,
-							!com.fasterxml.jackson.dataformat.yaml.snakeyaml,
 							*
 						</Import-Package>
 						<!--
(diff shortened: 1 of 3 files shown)


core/pom.xml
```diff
@@ -178,6 +177,12 @@
 	</build>
 	<dependencies>
 		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.15</version>
```
olivergondza: This is the version that used to be shaded by jackson: https://github.com/FasterXML/jackson-dataformat-yaml/blob/c69224917fd235732e51c8327f23af7f7038108f/pom.xml#L56.

auhlig:
Cool. Thanks a lot @olivergondza.

Issue #1094, opened 2017-10-05
NullPointerException due to HttpExecutorServiceImpl.execute returning null


pjdarton:
While testing the Jenkins openstack-cloud plugin, which uses openstack4 version 3.0.4j, I encountered a
```
java.lang.NullPointerException
	at org.openstack4j.openstack.internal.OSAuthenticator.authenticateV3(OSAuthenticator.java:156)
	at org.openstack4j.openstack.internal.OSAuthenticator.invoke(OSAuthenticator.java:70)
	at org.openstack4j.openstack.client.OSClientBuilder$ClientV3.authenticate(OSClientBuilder.java:165)
	at org.openstack4j.openstack.client.OSClientBuilder$ClientV3.authenticate(OSClientBuilder.java:128)
	at jenkins.plugins.openstack.compute.internal.Openstack.<init>
```

On investigation I found that this was caused by [`HttpExecutorServiceImpl`'s `execute` method](https://github.com/ContainX/openstack4j/blob/master/connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpExecutorServiceImpl.java#L28) catching Exception, calling `printStackTrace()` and then returning null (which the caller didn't check for) instead of throwing the Exception.
Further inspection revealed that the execute method is not documented as returning null, and there are multiple callers of this method which all assume that this method cannot return null.
One of these has to change - either we stop it from swallowing the exception and then returning null, or we cope with it returning null (personally, I think the first option is better).

 - To stop it from returning null,  [`execute`](https://github.com/ContainX/openstack4j/blob/3.0.4/connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpExecutorServiceImpl.java#L28) must be modified so that it throws an exception if it encounters one, e.g.
```java
    @Override
    public <R> HttpResponse execute(HttpRequest<R> request) {
        try {
            return invoke(request);
        }
        catch (ResponseException re) {
            throw re;
        }
        catch (RuntimeException e) {
            throw e;
        }
        catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
```

 - To cope with it returning null, we#d have to
   - Update the [API](https://github.com/ContainX/openstack4j/blob/3.0.4/core/src/main/java/org/openstack4j/core/transport/HttpExecutorService.java#L11) to document that this method can return null
[...]

PR #1095, opened 2017-10-05
[issue 1094] Fix NPEs caused by okhttp HttpExecutorServiceImpl

pjdarton:
org.openstack4j.connectors.okhttp.HttpExecutorServiceImpl.execute(...) no longer returns null.
If it encounters an exception, it now throws.
Previously it suppressed the original exception and returned null, causing downstream code to fail with a NullPointerException.

See https://github.com/ContainX/openstack4j/issues/1094 for details.


auhlig:
That was open for quite some time. Apologies. Thanks for fixing this @pjdarton. Can we close the corresponding issue?

pjdarton:
Yes; close it.

PR #1098, opened 2017-10-10
Update volume bootable status

lhwong:
Implement the following endpoint as per the Block Storage API document https://developer.openstack.org/api-ref/block-storage/v3/

```POST /v3/{project_id}/volumes/{volume_id}/action```
Update a volume’s bootable status

project_id | path | string | The UUID of the project in a multi-tenancy cloud.
volume_id | path | string | The UUID of the volume.
os-set_bootable | body | object | The os-set_bootable action.
bootable | body | boolean | Enables or disables the bootable attribute. You can boot an instance from a bootable volume.


vinodborole:
LGTM

auhlig:
Thanks a lot @lhwong.

Issue #1091, opened 2017-09-23
jackson suppress serializing properties with null values


Grim-lock:
https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/core/transport/ObjectMapperSingleton.java?utf8=%E2%9C%93#L25
Openstack4j does not serialize properties with null values, and this is a base function.

In octavia, PUT method of Pools:
"A JSON object specifying the session persistence for the pool or null for no session persistence"
pls see https://developer.openstack.org/api-ref/load-balancer/v2/#create-pool for more detail

Is there a easy way to config openstack4j to serialize a null value property?

@auhlig

auhlig:
Grrrrr. OpenStack. 
Can we we something like `@JsonInclude(Include.ALWAYS)` on the attribute @Grim-lock?

Grim-lock:
It works. I push a patch later

PR #1099, opened 2017-10-10
Fix octavia pool update session_persistence with value "null"

Grim-lock:
https://github.com/ContainX/openstack4j/issues/1091


auhlig:
💯  Thanks @Grim-lock

PR #1100, opened 2017-10-10
add rescue server status

hazard4j:
According to the OpenStack documentation there is a rescue status - https://developer.openstack.org/api-guide/compute/server_concepts.html

Thus I've added this status to Server.Status enum


hazard4j:
@vinodborole, @auhlig, could you please review this pull request?

There is only one little change but it's very vital for our project. 

Thank you in advance.

vinodborole:
LGTM

PR #1106, opened 2017-10-25
add detach method to blockstorage api

capitek-xun:
add detach method to blockstorage api


capitek-xun:
add detach method to blockstorage api

auhlig:
Thanks @capitek-xun

Issue #1117, opened 2017-11-16
Neutron Port creation does not support device_owner and device_id port properties


evgenyfedoruk:
When Building the port with a Port builder, it's possible to set the device owner and device id values via 
PortBuilder deviceOwner(String deviceOwner); method.

But, org.openstack4j.openstack.networking.domain.NeutronPortCreate class which builds the port create request body does not have those properies.

So port is created without the device owner and device id values.

auhlig:
Merged. Closing.

PR #1119, opened 2017-11-19
Adding device_owner and device_id support for Port

evgenyfedoruk:
Adding device_owner and device_id Port attributes
support to the port create model entity

Link to an issue: https://github.com/ContainX/openstack4j/issues/1117


auhlig:
Thanks @evgenyfedoruk!

PR #1120, opened 2017-11-20
Add cascade delete load balancer, it will delete all child objects of…

Grim-lock:
… the load balancer.


auhlig:
Thanks @Grim-lock

Issue #1122, opened 2017-11-22
Identity v3 -project().list() does not consider domain scope and fails


evgenyfedoruk:
The OSClientV3 identity().projects().list() fails to list projects.

Scenario:
1. Having domain DOMAIN1 and two projects, PROJECT1 and PROJECT2 under it.
2. Having user USER1 under the DOMAIN1 which has admin role on DOMAIN1.
3. Using identity V3 authentication with USER1 credentials and DOMAIN1 as domain identifier.
4. scoping the authentication to DOMAIN1.

```
OSClientV3 osv3 = OSFactory.builderV3()
	.withConfig(Config.newConfig().withSSLVerificationDisabled())
        .endpoint("http://10.205.124.40:5000/v3")
        .credentials("USER1", "PASSWORD" , Identifier.byId("d124580f9d8f41c69c7ed6c1bcf4ce5d"))
        .scopeToDomain(Identifier.byId("d124580f9d8f41c69c7ed6c1bcf4ce5d"))
        .authenticate();
```

5. Trying to get projects list (PROJECT1 and PROJECT2) of DOMAIN1

`
System.out.println("Projects:" + osv3.identity().projects().list());

`
6. Getting unauthorized error from keystone.

The issue is that Keystone expects domain_id value as a parameter in REST GET request"
http://10.205.124.40:5000/v3/projects?domain_id=d124580f9d8f41c69c7ed6c1bcf4ce5d

The project().list() just uses the basic projects URI for the request (http://10.205.124.40:5000/v3/projects)
It does not consider the domain_id inside the Authentication Token of the client.
The ProjectService class is not aware of the authentication token as far as I saw.
So, the requests are not considering the authentication scoping.

This issue can be fixed by adding a new API to the ProjectServiceImpl class along with the existing unparametrized list() API, like this:

```
    @Override
    public List<? extends Project> list() {
        return get(Projects.class, uri(PATH_PROJECTS)).execute().getList();
    }

    **@Override
    public List<? extends Project> list(String domainId) {
        return get(Projects.class, uri(PATH_PROJECTS)).param("domain_id", domainId).execute().getList();
    }**
```

I'm not sure this kind of fix is going along with OS4J standards.
If this fix is the right way to handle this issue, I can fix it.
I tested it and the new API is working properly.

Please share you thoughts on this.

PR #1123, opened 2017-11-22
Adding new ProjectService API for listing projects

evgenyfedoruk:
Adding new API for projects listing on certain domain
in ProjectService.

Keystone expects domain_id value in URL for domain scoped
authentication tokens to be authorized to list domain's projects.

OS4J ProjectService is not aware of the auth. token scope domain.

This patch allows to use projects list API
in a domain scope by specifying the domai_id as a parameter
in the keystone projects URL.

Link to an issue: #1122

### DIFF
--- a/core/src/main/java/org/openstack4j/api/identity/v3/ProjectService.java
+++ b/core/src/main/java/org/openstack4j/api/identity/v3/ProjectService.java
@@ -79,4 +79,11 @@ public interface ProjectService extends RestService {
 	 */
 	List<? extends Project> list();
 
+	/**
+	 * list all projects on the certain domain the current token has access to
+	 *
+	 * @return list of projects on a certain domain
+	 */
+	List<? extends Project> list(String domainId);
+
 }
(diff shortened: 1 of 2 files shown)


auhlig:
What are you trying to achieve? There is a `scopeToDomain(<domain is>)`

evgenyfedoruk:
I try to list domain's projects while having a domain scoped token.

I have domain "X" (id=d124580f9d8f41c69c7ed6c1bcf4ce5d)
I have a user on this domain called "lbaas-admin".
I do authentication with this user and scope the token to this domain.
Next, I want to list projects of this domain.
I expect the client to be aware of the domain scope.
Keystone expects to get the domain_id parameter on URL for GET request of projects
like: 
`
http://10.205.124.40:5000/v3/projects?domain_id=4398971401b54edb8d0127336b70961a
`
But the ProjectServiceImpl.list() method does not pass the domain_id parameter on the request

and I get unauthorized error.

```
OSClientV3 osv3 = OSFactory.builderV3()
	.withConfig(Config.newConfig().withSSLVerificationDisabled())
	.endpoint("http://10.205.124.40:5000/v3")
	.credentials("lbaas-admin", "password" , Identifier.byId("d124580f9d8f41c69c7ed6c1bcf4ce5d"))
	.scopeToDomain(Identifier.byId("d124580f9d8f41c69c7ed6c1bcf4ce5d"))
	.authenticate();            
System.out.println("LBaaS tenants:" + sv3.identity().projects().list());
```

The new API gets the domain id and adds it to the request.

I'm not sure this is the right way to solve this issue.
IMHO, in case of a domain scoped token, the domain id should be added by the ProjectServiceImpl class ,and for that, the ProjectServiceImpl should be aware of a token scope.


What do you think?

Issue #1135, opened 2017-12-18
Add support for abandon stack heat


akemr:
Heat REST API contains operations for both stack abandoning (_deletes a stack but leaves its resources intact, and returns data that describes the stack and its resources_) and adoption (_creates a stack from existing resources_)

See:
[https://developer.openstack.org/api-ref/orchestration/v1/#abandon-stack](url)
and
[https://developer.openstack.org/api-ref/orchestration/v1/#adopt-stack](url)

It would be nice to have support for these operations in OpenStack4J as well.

vinodborole:
@akemr we'd be happy if you can contribute

akemr:
Sure, I can contribute. Will open PR about it.

PR #1136, opened 2017-12-19
Heat service, adding support for stack abandon and adopt 

akemr:
Implements issue #1135. 

There are 2 new methods in `heat.StackService`: abandon and adopt. New interface `AdoptStackData` describes `adopt_stack_data` element, used as an input for stack adoption and as a return value for stack abandoning.

Basic unit tests are included. The code was tested against real Openstack instance as well.

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/heat/domain/HeatAdoptStackData.java
+++ b/core/src/main/java/org/openstack4j/openstack/heat/domain/HeatAdoptStackData.java
@@ -2,6 +2,7 @@ package org.openstack4j.openstack.heat.domain;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
 import java.util.Map;
 import org.openstack4j.model.heat.AdoptStackData;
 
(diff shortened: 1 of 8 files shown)


vinodborole:
@akemr thanks a lot for having this done so quickly. LGTM. @auhlig what do you think?

core/src/main/java/org/openstack4j/openstack/heat/domain/HeatAdoptStackData.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Override
+    public Map<String, Map<String, Object>> getResources() {
+        return resources;
+    }
+
+    @Override
+    public String toString() {
+        return "HeatAdoptStackData{" + "id=" + id + ", name=" + name + ", status=" + status + ", resources=" + resources + '}';
```
auhlig: I'd prefer
```
public String toString() {
MoreObjects.toStringHelper(this).omitNullValues()
  .add("id", id)
  .add(...)
  .toString();
}
```
What do you think @akemr?

akemr: Sure, no problem. My simple toString was just generated by IDE. 
@auhlig Should I do the change in my branch and add the commit to this pull request?

PR #1140, opened 2018-01-16
Add octavia error msg to ActionResponse

Grim-lock:
octavia error like this
{
    "debuginfo": null,
    "faultcode": "xxx",
    "faultstring": "xxx"
}

"debuginfo" can be NULL


auhlig:
LGTM. Thanks @Grim-lock

Issue #1124, opened 2017-11-27
openstack4j mistaken heat-cfn service as heat service endpoint URL


newinone:
The problem is:
openstack4j resolves service endpoint URL from the list of catalog. We are looking for “heat” service in particular.

- There is a Cloudformation service name is heat-cfn in the first place of catalog.
- openstack4j compare the ServiceType or service name start with service name: eg “heat”. And returns the first match which is heat-cfn
- the 1st issue (bug) is here:
https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/api/types/ServiceType.java#L52
- 2nd issue(feature request) Cloudformatio service is not in Service list.

Let me know if @junruh want me create a PR for proposed fix,

PR #1142, opened 2018-01-21
Fix for issue #1124:openstack4j mistaken heat-cfn service as heat ser…

newinone:
This is the proposed fix.
- The fix worked for our environment. 
- However it did change the previous matching behavior.
- I'm open to change the service pattern as long as the negative tests pass.


auhlig:
Thanks a lot @newinone. Can we test this somehow?

Issue #979, opened 2017-03-20
Disable Gateway problem


ayberkcal:
I have faced the same problem on https://github.com/ContainX/openstack4j/pull/599. "gateway_ip" property is not included in request json.
Subnet Builder Code:
```
Subnet subnet=getSubnetById(subnetId); //getting subnet from OS
SubnetBuilder builder=subnet.toBuilder(); //create builder from existed subnet
builder=builder.noGateway(); //disable subnet gateway
openStackService.client().networking().subnet().update(builder.build()); //send request to OpenStack API
```

Debug Logs:

> >> PUT /v2.0/subnets/9180b203-d4cb-4c92-9d3c-5b9958b10b82 HTTP/1.1 
> >> "{[\r][\n]" 
> >> "  "subnet" : {[\r][\n]" 
> >> "    "name" : "subnet name123",[\r][\n]" 
> >> "    "dns_nameservers" : [ ],[\r][\n]" 
> >> "    "allocation_pools" : [ {[\r][\n]" 
> >> "      "start" : "192.164.1.2",[\r][\n]" 
> >> "      "end" : "192.164.1.254"[\r][\n]" 
> >> "    } ],[\r][\n]" 
> >> "    "host_routes" : [ ][\r][\n]" 
> >> "  }[\r][\n]" 
> >> "}" 

As you can see "gateway_ip" property is not converted to json. I have debug code. Everything looks fine but samething went wrong when entity converting to json.

PR #1149, opened 2018-02-08
Fix for issue #979: Disable Gateway problem

ewhseo:
Fix for issue #979: Disable Gateway problem


auhlig:
Thanks @ewhseo and apologies for the later reply. This looks good. I hoped a unit test would have revealed that earlier. Can you extend the test? Would be much appreciated!

PR #1151, opened 2018-02-11
Make sure the number of okhttp CLOSE_WAIT connections is always manageable

olivergondza:
I have observed in production 2500ish openstack connections in `CLOSE_WAIT` state that took several minutes to get properly closed (by `ConnectionPool#cleanup()`). I turned out every interaction with openstack creates separate `OkHttpClient` instance which is discouraged so dedicated auto-cleaned connection pools have no chance to have any positive effect. I am not convinced it is wise to share the client/connectionpool for the `HttpExecutorServiceImpl` so I am settling with the eager resource cleanup.

I have retested this change via jenkinsci/openstack-cloud-plugin and the interaction that used to create 100+ connection that took a while to be released, this hardly ever create more than 5 at the same time.

### DIFF
--- a/connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpCommand.java
+++ b/connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpCommand.java
@@ -12,6 +12,7 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import okhttp3.Call;
+import okhttp3.ConnectionPool;
 import okhttp3.Interceptor;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
@@ -90,10 +91,25 @@ public final class HttpCommand<R> {
         if (HttpLoggingFilter.isLoggingEnabled()) {
             okHttpClientBuilder.addInterceptor(new LoggingInterceptor());
         }
+        okHttpClientBuilder.connectionPool(getConnectionPool());
         client = okHttpClientBuilder.build();
         clientReq = new Request.Builder();
         populateHeaders(request);
         populateQueryParams(request);
+    }
+
+    /**
+     * Create ConnectionPool optimized for short lived client with little chance to reuse connections.
+     */
+    private ConnectionPool getConnectionPool() {
+        int maxIdleConnections = 0;
+        // OkHttp creates "OkHttp ConnectionPool" thread per every ConnectionPool created to mange its connections. It
+        // lives as long as the last connection made through it + its keep alive timeout. By default that it 5 min which
+        // makes no sense for openstack4j since the connections or pools are not reused (after HttpCommand completion,
+        // at least). Setting strict keepAlive duration here so the connections and threads does not hang around longer
+        // than necessary.
+        int keepAliveDuration = 500;
+        return new ConnectionPool(maxIdleConnections, keepAliveDuration, TimeUnit.MILLISECONDS);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


olivergondza:
Any opinions?

vinodborole:
Thanks @olivergondza for the change. LGTM @auhlig ?

olivergondza:
Thanks, is there anything I can do to have this released promptly? I am asking if workaround is worth devising...

auhlig:
Many thanks @olivergondza and apologies for the delay.
This looks more like a workaround. There shouldn't be a need to clean up from our side, so we would want this fixed in okhttp, not?

olivergondza:
Please hold on with merging, I have discovered it creates per-connection threads as well.

---

Well, yes this is a workaround BUT I doubt there is something for them to fix. The thing is that openstack4j is using the client in a different way than recommended by OKHttp team. According to javadoc, the client is optimized for a single per-app client instance for all HTTP calls to go through. The default keepalive of 5 minutes and connection/thread pool with no size limit make lot more sense for such use-case. It does not guarantee the size of open connections / running threads is always be in sane numbers (which I would be interested in anyway) BUT those connections will be actually reused. Since openstack4j create new client per `HttpCommand`, these assumptions bring no real benefits for us as and we have large number of connection/threads around for no gain.

In the PR, I am configuring the client not to expect connection reuse / clean the resources eagerly reducing the resource waste created. An alternative approach would be to use long-lived client for okhttp though I do not know if there is some reasonable scope that would hold it. Running the single client per-JVM (static field lazy-initialized field in `HttpCommand`?) sounds fragile to me as for my use-case it would mean the single instance should work fine for weeks or months of sustained load. I am not saying okhttp can not do that it is just our instances of it was extremely short-lived for now and I am afraid changing it to extremely long-lived ones would just open a different can of worms which is about the last think I am after.

@auhlig, does that make sense?

auhlig:
Thanks for the explanation. 
So the way we're using okhttp is preferable despite it's not the one that's recommended?

olivergondza:
In the end, I found simpler way to constrain both network connections and ConnectionPool bookkeeping threads. I kept the commit reverted for the review, will squash it once we agree the is the way to go.

olivergondza:
@auhlig, now I have successfully re-tested this through the Jenkins plugin and confirmed the old version misbehaves consistently with my observations (TCP connections open for 5 minutes after used, threads exists for 5+1 minutes) as well as the new implementation disposes connections almost instantly and threads after 1 minute of being idly in the threadpool.

WRT our preference, openstack4j have been recreating clients per command for a long time and we know it works reasonable well. There ware a problem we managed to reproduce and verify the fix. Whether the shared long-running client approach will work us equally well is a question that is more challenging to answer as we have no past data (AFAICT) and verifying it would require months of soak testing. IMO this is what openstack4j should prefer.

olivergondza:
> When will the keepAliveDuration kick in? It's just for idle connection and longer running things (file up-/downloads) won't be affected, correct?

That is my understanding. I have just retested by downloading 2.8GB image. It took 48 minutes, created one RCP connection and one thread, both disappeared once successfully downloaded.

auhlig:
Cool. Thanks for testing. I’ll merge later 
Might other connectors be affected as well?

olivergondza:
In theory, they can. I find the idea of one shared client instance per whole application a bit strange myself but if authors of the other connection libraries think alike they might be affected too.

olivergondza:
@auhlig, do you have an idea of when this can get released? I am asking to see if it is worth releasing openstack-cloud with snapshot dep in it. Thanks!

Issue #1152, opened 2018-02-21
Add support for Application Credentials in the Identity V3 api


mariusleu:
In 3.10 there was introduced the Application Credentials API.
https://developer.openstack.org/api-ref/identity/v3/#application-credentials
It would be nice to have that included in Openstack4j.

konetzed:
Any progress on Application Credentials support for opensatck4j?

yasarutkuu:
any progress?

PR #1153, opened 2018-02-21
Issue #1152 - added the Application Credential API

mariusleu:


vinodborole:
@mariusleu can you please fix the checks that are failing, we need to cut out a new release

PR #1162, opened 2018-03-02
Consider the SSLContext of an HttpRequest's Config object when using the Resteasy connector

### LINKED ISSUE #1161: Resteasy connector ignores SSLContext
The Resteasy connector does not check whether the HttpRequest's Config object has an SSLContext.
Other connectors perform this check e.g. the [httpclient connector](https://github.com/ContainX/openstack4j/blob/8fb99200d16a64888060c8c308be04a2b602115e/connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpClientFactory.java#L76).

Therefore something like this
```
Config config = Config.newConfig();
config.withSSLContext(sslContext);
OSClient os = OSFactory.builderV3().withConfig(config)
```
does not work as expected.

ThomasBri:
The SSLContext of an HttpRequest was not passed to the HttpClientBuilder in the ApacheHttpClientEngine. This closes #1161.


auhlig:
Thanks a lot @ThomasBri

Issue #1165, opened 2018-03-12
Httpclient connector deadlock


mnitchev:
Greetings,

We noticed a bug similar to #400  in the reauthentication logic in `openstack4j-httpclient` connector. This is the code in question in  [HttpExecutorServiceImpl](https://github.com/dbantchovski/openstack4j/blob/af9c2db7c01665647c76c2df5ade46883c707e4e/connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpExecutorServiceImpl.java#L64-L80):
```
private <R> HttpResponse invokeRequest(HttpCommand<R> command) throws Exception {
    CloseableHttpResponse response = command.execute();

    if (command.getRetries() == 0 && response.getStatusLine().getStatusCode() == 401 && !command.getRequest().getHeaders().containsKey(ClientConstants.HEADER_OS4J_AUTH))
    {
        try
        {
            OSAuthenticator.reAuthenticate();
            command.getRequest().getHeaders().put(ClientConstants.HEADER_X_AUTH_TOKEN, OSClientSession.getCurrent().getTokenId());
            return invokeRequest(command.incrementRetriesAndReturn());
        } finally {
            response.close();
        }
    }

    return HttpResponseImpl.wrap(response);
}
```
The problem is that the connection created on [this](https://github.com/dbantchovski/openstack4j/blob/master/connectors/httpclient/src/main/java/org/openstack4j/connectors/httpclient/HttpExecutorServiceImpl.java#L65) line is not being closed before recursively calling the same method. The code in the finally block is executed **after** the recursive invocation completes which means that a new connection will be requested without closing the currently active one first.

By default Apache's HttpClient only allows 2 concurrent connections for one route. It is possible that two concurrent requests to the same route could fail to authenticate at the same time which will trigger two reauthentications and two recursive invocations of invokeRequest.
That unsuccessfully tries to create two new connections for the same route because the quota for allowed connections would be already filled with the initial connections. That causes a deadlock.

Here's the example workflow that caused our problem:

-> GET route1
<- returns 401 (token has expired)
-> POST route2 (token reauthentication) (1)

-> GET route1
<- return 401 (token has expired)
-> POST route2 (token reauthentication) (2)

(1) <- returns 201
-> GET route 1 (this blocks because there are already two active connections for the same route)

(2) <- return 201
-> GET route 1 (this blocks because there are already two active connections for the same route)

[...]

PR #1166, opened 2018-03-13
Fix issue #1165.

gdankov:
Moved the return method after the finally block thus guaranteeing that the initial connection is always closed before the recursive invocation.


auhlig:
Thanks a million @gdankov!

PR #1183, opened 2018-04-19
member of extra field in KeystoneProject(v3) should be Object Type

caitsithx:
In case some json attribute is not string typed in the payload, the mapping will fail.


JohannesRudolph:
I agree this would be a sensible change to improve version compatibility / resiliency without requiring an immediate update of OS4j itself (e.g. see the "tags" field added in OpenStack Queens which is a `string[]` and thus fails to deserialize).

Issue #1180, opened 2018-04-14
Unchecked Exception is not thrown for http status 404


pravinrajr9:
The execute method of BaseOpenStackService class does not throw exception for http response status 404.
Instead an error message is logged. It would be better if it had thrown exception for illegal access. So that the one who is using this client would understand the cause more faster.  Currently it is returning null as response for any http response with 404 status.
Is it left for logging for any use case or it has to be fixed?

I would also be glad to fix and contribute for openstack4j

pravinrajr9:
@gondor , @auhlig I did not get purpose of not throwing an exception for response with http 404 status code. I would be glad to fix this, if it is expected to throw an exception

vinodborole:
Throwing exception or checking specific status code is not required with the core library as it will not be generic, the Response Object returned is sufficient for the caller to make use of it they way it wants.

pravinrajr9:
@vinodborole Wondering why there is no exception thrown for only response having http error code 404.
While other error codes less then 500 and not equal to 404, an exception is thrown. The response object returned is null and again we need to make an extra call to know the actual cause.

vinodborole:
its handled here - https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/core/transport/HttpExceptionHandler.java

Please go through the implementation for handling exceptions.

pravinrajr9:
@vinodborole I agree but in the HttpEntityHandler  line no 43 returns true so it never reach the line                ` throw mapException(response.getStatusMessage(), response.getStatus());`  where it calls mapExection in HttpExceptionHandler.java 
 
So it is returning null instead of throwing an exception.

pravinrajr9:
@vinodborole  In method handle404 below

```
    private static <T> Handle<T> handle404(Handle<T> handle) {
        if (handle.getResponse().getStatus() == 404) {

            if (ListType.class.isAssignableFrom(handle.getReturnType())) {
                try {
                    return handle.complete(handle.getReturnType().newInstance());
                } catch (InstantiationException e) {
                    LOG.error(e.getMessage(), e);
                } catch (IllegalAccessException e) {
                    LOG.error(e.getMessage(), e);
                }
            }

            if (handle.getReturnType() != ActionResponse.class) {
                return handle.complete(null);
            }
        }

        return handle.continueHandling();
    }
```

I think, not sure much, there might be problem, in the if condition and  comparing for ActionResponse class != handle.getReturnType(). 
[...]

PR #1188, opened 2018-05-03
Fix: To throw unchecked exception for HTTP status 404

pravinrajr9:
For unchecked exceptions handling in handle404 method,
in most of the cases the handle.getReturnType is usually not
of type action response class. If it is of type ActionResponse class
the code has been changed to successfully return ActionResponse object.
It is good throw all unchecked exceptions if handle.getReturnType is not equal to
Action Response.


pravinrajr9:
To Solve issue https://github.com/ContainX/openstack4j/issues/1180

PR #1191, opened 2018-05-24
Add support for “tags” of project which introduced in KeyStone v3.9

### LINKED ISSUE #1163: Project tags introduced in Identity API v3.9 are not supported
In the Openstack Identity API v3.9 there was introduced the new "tags" field in the Project resource.
It seems Openstack4j is not compatible with it, causing the following JsonMappingException:

```
javax.ws.rs.ProcessingException: com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.lang.String out of START_ARRAY token
 at [Source: java.io.ByteArrayInputStream@413ecd76; line: 1, column: 179] (through reference chain: org.openstack4j.openstack.identity.v3.domain.KeystoneProject["tags"])
```

baseman79:
Found the issue when creating project against OpenStack Queens

Fixes #1163

### DIFF
--- a/core/src/main/java/org/openstack4j/openstack/identity/v3/domain/KeystoneProject.java
+++ b/core/src/main/java/org/openstack4j/openstack/identity/v3/domain/KeystoneProject.java
@@ -1,17 +1,23 @@
 package org.openstack4j.openstack.identity.v3.domain;
 
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.fasterxml.jackson.annotation.*;
 import org.openstack4j.model.identity.v3.Domain;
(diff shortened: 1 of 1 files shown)


vadsiva:
The added support is very important because it allows you to use the latest releases since pike! Its presence in the upcoming release is very desirable.

B0bSapp:
Actually we've  already faced this error after updating our openstack environment. So this fix is really important for us

vinodborole:
Please feel free to contribute and raise a PR.

vadsiva:
@vinodborole  PR (https://github.com/ContainX/openstack4j/pull/1219, https://github.com/ContainX/openstack4j/pull/1191) is created for the second time. What's next?

PR #1226, opened 2018-11-13
Implement listCachedImages() for Glance V2 Images

jyothisaroja:
This pull request implements the listCachedImages() API to list the cached images for Glance v2 images.
This API is available in glance v1 but is missing in v2 version. Hence implemented the same for V2.


biogerm:
lgtm

PR #1274, opened 2019-10-14
Use same connection pool for all clients per OKHttp recommendations

sbrattla:
The OKHttp documentation recommends the following usage pattern:

> The best practice in OkHttp 3 is to create a single OkHttpClient instance and share it throughout the application. Requests that needs a customized client should call OkHttpClient.newBuilder() on that shared instance. This allows customization without the drawbacks of separate connection pools.

The current `HttpCommand` implementation for OKHttp creates a new `OkHttpClient` for each request, and subsequently a new connection pool for each request. This may not be an big issue since the connection pool is set to allow 0 connections and also to evict connections after 500 ms. Nevertheless, the overhead of initializing a new connection pool and related threads for each and every http request may not be negligible for a large number of http requests.

This pull request attempts to address this by providing a singleton client available to all requests. This singleton client is not the client used for the actual http requests, but serves as a "template" client. We use `client.newBuilder()` to create a shallow copy of the singleton client. This will reuse the existing connection pool, but still enable us to fully customize each http request.

The singleton OkHttpClient will be instantiated in the `initialize` method of `HttpCommand`.

### DIFF
--- a/connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpCommand.java
+++ b/connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpCommand.java
@@ -41,9 +41,10 @@ import com.google.common.io.ByteStreams;
 public final class HttpCommand<R> {
 
     private static final Logger LOG = LoggerFactory.getLogger(HttpCommand.class);
+    private static OkHttpClient client;
 
-    private HttpRequest<R> request;
-    private OkHttpClient client;
+    private final HttpRequest<R> request;
+    private OkHttpClient reqClient;
     private Request.Builder clientReq;
     private int retries;
 
@@ -63,7 +64,15 @@ public final class HttpCommand<R> {
     }
 
     private void initialize() {
-        OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder();
+        if (client == null) {
+            synchronized (HttpCommand.class) {
+                if (client == null) {
+                    client = getClient(request.getConfig());
+                }
+            }
+        }
+
+        OkHttpClient.Builder okHttpClientBuilder = client.newBuilder();
         Config config = request.getConfig();
 
         if (config.getProxy() != null) {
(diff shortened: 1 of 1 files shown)


connectors/okhttp/src/main/java/org/openstack4j/connectors/okhttp/HttpCommand.java
```diff
@@ -101,15 +110,23 @@ private void initialize() {
...
+    private static OkHttpClient getClient(Config config) {
         int maxIdleConnections = 0;
         // OkHttp creates "OkHttp ConnectionPool" thread per every ConnectionPool created to mange its connections. It
         // lives as long as the last connection made through it + its keep alive timeout. By default that it 5 min which
         // makes no sense for openstack4j since the connections or pools are not reused (after HttpCommand completion,
         // at least). Setting strict keepAlive duration here so the connections and threads does not hang around longer
         // than necessary.
         int keepAliveDuration = 500;
-        return new ConnectionPool(maxIdleConnections, keepAliveDuration, TimeUnit.MILLISECONDS);
+        ConnectionPool cp = new ConnectionPool(maxIdleConnections, keepAliveDuration, TimeUnit.MILLISECONDS);
```
olivergondza: Author of the original workaround here. I like the approach of rectifying the misuse of OkHttp client, though I belive we have to rethink the keepalive duration and the size of the pool. The only reason this was ever tempered with was the numerous `OkHttpClient` instances took minutes to purge the threads and connections causing resource waste. I guess we can get back to OkHttp's defaults here as we are using the (single) client the was it was meant to be.

olivergondza: I do not think we need any of this now: https://github.com/ContainX/openstack4j/commit/dc97e6733a0d1151cce69dd9f453a6b501faf323#diff-e08fdcb99d047ca41bbe3fe37e41802f

PR #1296, opened 2020-02-17
Point users to new upstream repository: openstack4j/openstack4j

olivergondza:
We are continuing the project development in [openstack4j/openstack4j](https://github.com/openstack4j/openstack4j/) for the lack of traction here[1]. I am filing this to help users to easily locate the active upstream.

@auhlig, @gondor, please have a look. I would appreciate if you would integrate this.

[1] https://groups.google.com/forum/#!topic/openstack4j/pY2eee760QQ


auhlig (approved):

auhlig:
Does that make sense to you @gondor?

gondor:
@olivergondza let's keep this in the original repo since it has the larger user base.  Let me know what permissions and access you want so we can keep It there and give you the flexibility to be the main maintainer

gondor:
@auhlig ^^ for visibility since we were emailing each other about this

olivergondza:
@gondor, thanks for getting back to us and offering the possibility to continue the development here!

I would like to keep the vision of building a community around the project and distributing the permission across several people to eliminate the bus factor and to have a healthy developer ecosystem. Also, if we move the development back here, I find it important to merge back openstack4j/openstack4j and openstack4j/openstack4j.github.io including the releases produced there (so the version numbers will not clash).

Speaking of permissions needed, these are the ones I can think of:

- Becoming an owner of the github repos openstack4j and openstack4j.com (probably undesirable for the whole group) so we can manage the code, teams, integrations, etc.
- Administrative permissions to the google discussion group.
- Permissions to release bits to maven central.
- We also need a long term commitment to maintain and fund the domain registration. I would be happy to take that, as well.

With all those announcements I published lately about separating the project to openstack4j github group, it might actually be smoother to just stick to that one. I believe with explicit control of the webpage and the source repo, we can redirect the users quite effectively - so I am offering the possibility of just finishing the move there while keeping you guys as maintainers in case you are interested. Though, I would understand if you feel strongly for ContainX group.

Issue #1328, opened 2020-11-09
Join us to solve any problems of this project


bboyHan:
fork link: https://github.com/bboyHan/openstack4j

We are willing to spend more time to solve any problems of this project, including bug fixing and function extension.  Welcome to communicate here, and welcome to join us to complete it.

PR #1329, opened 2020-11-09
implement extention - 4.0 plus. join us

bboyHan:
#1328 

changelog：
1. add neutron class time field about [network/subnet/port];
2. optimize some code to replace the Deprecated methods and fields;
3. add neutron api implement extention about network-ip-availabilities;(link: https://docs.openstack.org/api-ref/network/v2/#network-ip-availability-and-usage-stats);
4. add (Neutron) resource tags operation. (link: https://docs.openstack.org/api-ref/network/v2/#standard-attributes-tag-extension);


bboyHan:
5. add qos policy extension; (link: https://docs.openstack.org/api-ref/network/v2/#qos-policies-qos)

olivergondza:
@bboyHan, you seem to have filed this in an incorrect project.

Also, never mix unrelated changes in a single PR. And careful with code reformats, it is best avoided in distributed development. In openstack4j/openstack4j we would have to do so, but in a coordinated way: https://github.com/openstack4j/openstack4j/issues/21

